PR #3, opened 2015-04-24
unbind resize and readystatechange events in componentWillUnmount

stefanduberg:
This PR cleans up the resize and readystatechange eventlisteners added in componentDidMount.

Currently the `setDimensions` method throws `Uncaught TypeError: Cannot read property 'offsetWidth' of null` if you resize the browser window after this component have been unmounted.


kenwheeler:
Awesome! Thanks!

PR #12, opened 2015-06-24
Update package.json

MozMorris:
So this components shows up on http://react.parts/web


MozMorris:
Of course, you may not want that but I had to track this down through a GitHub search and your original slider.

PR #21, opened 2015-09-02
Add npmignore and remove compiled assets from git

amsardesai:
This PR does a few things:
- Adds a `.npmignore` the repository that makes sure things that aren't needed in the final build (like tests and uncompiled source files) are not in the final published package that we all download when we do `npm install nuka-carousel` on our projects.
- Removes compiled files from repository. These files are not needed in a code repository, they belong in the published package.
- Adds a `prepublish` task so that when you do `npm publish` the build task runs automatically. Kind of like integrating npm with babel.
- Removes `npm-debug.log` from repository.


ryan-roemer:
Thanks for the help!

We've already switch to an `.npmignore` whitelist in `master` and we're moving to fully honed down `npm version` workflow to finally do this all right, so I'm closing here.

PR #27, opened 2015-09-29
Fix incorrect event variable name in onTouchStart

Merri:
Fixes `event` to `e` and removes unused `event` argument.

Current code before this change throws errors.


Merri:
It would be nice to get this merged asap as this bug causes unnecessary noise to anyone with error logging, even if the functional issues caused by this issue aren't very notable.

kenwheeler:
Thanks!

Merri:
Thanks as well!

PR #36, opened 2015-11-02
Take cell spacing into account during touchMove

### LINKED ISSUE #34: Using cellSpacing causes touch slide to jump
If you add a cellSpacing swiping between slides become increasingly jumpy as you get longer into the slide deck. This happens on iOS (Android too perhaps?), but works well on Desktop as far as I can tell. I guess there might be some thouch calculation that fails to properly adjust for cellSpacing?

Can be replicated by adding cellSpacing={20} to the live example on the nuke-carousel site using any touch device.

I will try to see if I can find the issue and submit a fix PR in the coming days. If you can give me any pointers I'd appreciate it.

eldh:
Fix #34.

Add cellSpacing to the top/left calculations in the onTouchMove handler.

### DIFF
--- a/src/carousel.js
+++ b/src/carousel.js
@@ -191,8 +191,8 @@ const Carousel = React.createClass({
         }
 
         self.setState({
-          left: self.props.vertical ? 0 : (self.state.slideWidth * self.state.currentSlide + (self.touchObject.length * self.touchObject.direction)) * -1,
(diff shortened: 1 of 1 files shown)


eldh:
Not sure if this is fixed or not, but not relevant for me, so closing.

PR #38, opened 2015-11-03
Updating decorators proptypes to be more specific.

eriklharper:
Thought this would be a good way to provide some built-in documentation and prop validation at the same time for the `decorators` prop.


eriklharper:
thanks @kenwheeler !

PR #41, opened 2015-11-26
Use translate3d to animate list styles

eldh:
The performance when animating left and top properties was really bad in some cases, resulting in a lot of lag. 

This PR changes to animate the translate3d property instead, which is more performant since the browser doesn’t need to repaint on every frame.

Tests are also updated accordingly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,8 +39,8 @@
     "karma-webpack": "^1.7.0",
     "mocha": "^2.3.3",
     "phantomjs": "^1.9.18",
-    "react": "^0.14.0",
-    "react-dom": "^0.14.0",
+    "react": "^0.14.3",
+    "react-dom": "^0.14.3",
     "react-hot-loader": "^1.3.0",
     "sinon": "^1.17.1",
     "sinon-chai": "^2.8.0",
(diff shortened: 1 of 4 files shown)


reel:
+1

package.json
```diff
@@ -49,6 +49,10 @@
     "webpack": "^1.12.2",
     "webpack-dev-server": "^1.12.0"
   },
+  "peerDependencies": {
```
kenwheeler: why add peer deps here?

eldh: It might be a bit unrelated to the main purpose of this PR, but I think it's good to specify which version of React the package requires in `peerDependencies` so consumers are properly warned if wrong version of React is in place. If it was added to `dependencies` instead you'd risk loading two versions of React and just having it in `devDependencies` would not give consumers any info about that.

kenwheeler:
Can we make sure that we are supporting all the prefixes and falling back to left here?

eldh:
No reason to add mozTransform I think. msTransform would add ie9 compatibility, so should do that. Should this lib support IE < 9?

dmitry-sher:
this is very-very needed PR for us

kenwheeler:
Ok I'll review it locally tomorrow and get a version cut if everything checks out

dmitry-sher:
Here's what i did for our local testing. Works as a charm in webkit-based browsers and desktop/mobile safari.

https://github.com/dmitry-sher/nuka-carousel/commit/e742600e535692c2af57867aeb5a4db963dcc612

eldh:
Just let me know if I need to make any more updates to this. Would also love to see it landed!

PR #49, opened 2016-02-02
add contribution docs

djkirby:
I had some trouble getting started with the code base, figured some contribution docs would be useful.


kenwheeler:
Thanks!

PR #53, opened 2016-02-05
enable conditional rendering slide inside

hanwencheng:
which will enable the sentences like the followings, otherwise the carousel will render the following sentences into a blank slide.

```
        const images = []; //which could be a list of files later.
        <Carousel
          ref="carousel"
          data={this.setCarouselData.bind(this, 'carousel')}>
          <img src="http://placehold.it/1000x400&text=slide1"/>
          {images.length >= 1 && images.map(image => <img src={window.URL.createObjectURL(image)}/>)}
        </Carousel>
```

### DIFF
--- a/src/carousel.js
+++ b/src/carousel.js
@@ -135,7 +135,7 @@ const Carousel = React.createClass({
 
   render() {
     var self = this;
-    var children = React.Children.count(this.props.children) > 1 ? this.formatChildren(this.props.children) : this.props.children;
+    var children = this.countChildren(this.props.children) > 1 ? this.formatChildren(this.props.children) : this.props.children;
     return (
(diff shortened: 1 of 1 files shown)


kenwheeler:
I'm not sure I understand what this is doing?

hanwencheng:
`{imageUrls.length >= 1 && imageUrls.map(imageUrl=> `

This behavior the same like 

```
if(imageUrls.length >=1 ){
  // now render the image DOM element
  imageUrls.map(imageUrl=> <img src={imageUrl}/>
}

```

Without this change, when the imageUrls is an empty array, it render the code above as {false} in its props.children, which would be a blank slider. For such code style, this change is needed, but whether to adopt this is according to personal taste

PR #65, opened 2016-03-14
Fix Vertical Scrolling on Touch Devices

tinystride:
The `length` calculation in `onTouchMove` was not checking for the presence of the `vertical` property. 

This caused a significant bug when scrolling vertically on touch devices when using `vertical: true`: a user has to swipe **horizontally** to make the carousel transition **vertically**.

This PR fixes the scrolling bug by checking for `vertical` property in `onTouchMove`, and calculating the touch's `length` with `pageY` and `startY` rather than `pageX` and `startX`. A similar conditional already exists for mouse events: https://github.com/kenwheeler/nuka-carousel/blob/master/src/carousel.js#L267-L268, so the behavior there is correct.

### DIFF
--- a/lib/carousel.js
+++ b/lib/carousel.js
@@ -209,12 +209,14 @@ var Carousel = _react2['default'].createClass({
           e.preventDefault();
         }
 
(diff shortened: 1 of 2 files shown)


kenwheeler:
Thanks!

PR #68, opened 2016-03-29
Refactor props.children.count to only include valid react elements

lauriejones:
Hi Ken -- Awesome carousel. Thanks very much for open-sourcing it.

I am looking to use it in a project at work to swipe between a few products based on some filtering by a user.

I had a small issue with how the slideCount is calculated.

Our use case should show 3 products based on the user's choices. If our search returned less than 3 products we were displaying a message in place of the missing third product. If not, this just returned `null`. In this situation Nuka Carousel was adding an extra slide to the carousel for `null` so that we were always showing 4 slides instead of the 3 we should have been.

This solution checks that each of the children are valid elements before adding them to the count.

I have adjusted the example to show the problem this PR solves. If you want to try the current version with an invalid element you'll see that it allows you to swipe to an empty slide.

Thanks again,
Laurie

### DIFF
--- a/src/carousel.js
+++ b/src/carousel.js
@@ -580,14 +580,14 @@ const Carousel = React.createClass({
     var listWidth = this.state.slideWidth * React.Children.toArray(this.props.children).filter(c => React.isValidElement(c)).length;
     var spacingOffset = this.props.cellSpacing * React.Children.toArray(this.props.children).filter(c => React.isValidElement(c)).length;
     var transform = 'translate3d(' +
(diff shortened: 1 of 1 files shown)


kenwheeler:
This PR looks awesome! can you remove the example from the demo and rebase and I'll get this merged?

lauriejones:
Thanks Ken! Let me know if there is anything left to address.

mariano-formidable:
Closing this PR because it's too old to merge, but addressing this issue in our current build under PR #379

PR #81, opened 2016-04-16
'visibility' should use 'inherit', not 'visible'

### LINKED ISSUE #80: "visibility" should be "inherit", not "visible", to avoid overriding "hidden" on parent containers
The `visibility: visible;` style makes an element visible regardless of the parent containers' `visibility: hidden;`.

The `visibility: inherit;` should be used as the opposite to `hidden` here:
https://github.com/kenwheeler/nuka-carousel/blob/111b6f16f8304b3c9fea17f192d087de264ec603/src/carousel.js#L643
```
visibility: this.state.slideWidth ? 'visible' : 'hidden'
```

sompylasar:
Fixes #80.

The `visibility: visible;` style makes an element visible regardless of the parent containers' `visibility: hidden;`.
The `visibility: inherit;` should be used in this component as the opposite to `hidden` to allow parent containers to hide this element via `visibility: hidden;`.

Reference: https://jakearchibald.com/2014/visible-undoes-hidden/

### DIFF
--- a/src/carousel.js
+++ b/src/carousel.js
@@ -640,7 +640,7 @@ const Carousel = React.createClass({
       height: 'auto',
       boxSizing: 'border-box',
       MozBoxSizing: 'border-box',
-      visibility: this.state.slideWidth ? 'visible' : 'hidden'
+      visibility: this.state.slideWidth ? 'inherit' : 'hidden'
     }
   },
 


benbayard:
To maintain backwards compatibility it might be better to have `visibilityInherit` as a prop.

mariano-formidable:
This PR is too old to merge in, but the fix has been addressed in PR #380 in our current build.

PR #85, opened 2016-05-01
Add additional class to all visible slides

patrickgalbraith:
This pull request adds a class to all visible slides. 

I needed a way to change the appearance of the currently visible slides. In my case I am setting the `max-height` to something other than `0px` for visible slides to so the slide height is automatic.

### DIFF
--- a/src/carousel.js
+++ b/src/carousel.js
@@ -483,8 +483,13 @@ const Carousel = React.createClass({
 
   formatChildren(children) {
     var self = this;
+
+    const { currentSlide } = this.state;
+    const { slidesToShow } = this.props;
+
     return React.Children.map(children, function(child, index) {
-      return <li className="slider-slide" style={self.getSlideStyles()} key={index}>{child}</li>
+      let visible = index >= currentSlide && index < currentSlide + slidesToShow;
+      return <li className={'slider-slide' + (visible ? ' slide-visible' : '')} style={self.getSlideStyles()} key={index}>{child}</li>
     });
   },
 


CharlyJazz:
I think maybe this feature is not necessary. Because you can manage the state of the slider from a parent component and thus add an "Active" class to the active slider

``` javascript
import React, { Component } from 'react'
import Carousel from 'nuka-carousel';
import { PagingDots } from '../PagingDots';
import './styles.scss'

export default class index extends Component {
  state = {
    slideIndex: 0
  };

  static defaultProps = {
    data: [
      {....},
      {....},
      {....}
    ]
  }

  render() {
    return (
      <div className='Slider'>
        <Carousel
          autoplay
          renderBottomCenterControls={props => <PagingDots {...props} />}
          renderCenterLeftControls={null}
          renderCenterRightControls={null}
          heightMode='first'
          autoplay={true}
          slideIndex={this.state.slideIndex}
          afterSlide={slideIndex => this.setState({ slideIndex })}
          easing='easeQuad'
        >
          {
            this.props.data.map((i, index) => (
              <div className='Slider-Item'>
                <img
                  src={i.src}
                  alt={i.title}
                  onLoad={() => { window.dispatchEvent(new Event('resize')); }}
                />
                <div className={`Slider-Caption ${this.state.slideIndex === index ? '--Active' : ''}`}>
                  <div>
                    <h1>{i.title}</h1>
                    <p>{i.description}</p>
                  </div>
                </div>
              </div>
            ))
          }
        </Carousel>
      </div>
    )
  }
}
```

It would be excellent if there was a section in the documentation describing how to do this.

patrickgalbraith:
@CharlyJazz I'm sure a lot has changed in the > 2 years since I opened this pull request.

I agree it is likely no longer necessary.

sarmeyer:
heightMode = current will also take care of this use case

PR #95, opened 2016-06-06
Revert "adding autoplay"

kenwheeler:
Reverts kenwheeler/nuka-carousel#90


rochestb:
Ken did this break?

kenwheeler:
So generally speaking, the autoplay interval should pause and reset during interaction. Also, it just gets to the end and stops. We should go back to the first slide. I should also probably implement wraparound.

PR #101, opened 2016-06-13
Issue #100: removed autoplay interval on carousel unmount

### LINKED ISSUE #100: Bug: carousel component tries to call setState of unmounted component because of autoplay set interval
We should to remove autoplay calls interval.
For this we need to call this.stopAutoplay() in componentWillUnmount().

IpShot:
Fixes #100


kenwheeler:
Thanks!

PR #109, opened 2016-06-27
Set button type on controls to 'button'

jonnywyatt:
This prevents 2 types of unwanted behaviour when the prev/next/bullet buttons are clicked -
1 - the page won't jump to the top
2 - any surrounding form elements won't submit


mariano-formidable:
In the latest version of `nuka-carousel` (4.3.2), all the built-in slider controls (next/prev/paging dots) are now `button` types.

Issue #105, opened 2016-06-20
Hardcoded slide height if no slides


krom:
Use this.props.initialSlideHeight instead 100 at https://github.com/kenwheeler/nuka-carousel/blob/master/src/carousel.js#L660

If Slides added dynamically initial height will be 100 until page size changed

kenwheeler:
want to PR that?

Antherkiv:
Hello I facing a similar situation, I had the component as a header coupling with react-router and I fetch the image on client (universal rendering), so the component never goes well, if I put the little hack suggested by someone here, the component it's 0 until the images are loaded, I don't know if you had a good idea to handle, thanks

ryan-roemer:
Is this still an issue on `nuka-carousel@3.1.0+` now that we've landed https://github.com/FormidableLabs/nuka-carousel/pull/291?

ncuillery:
@ryan-roemer Yes it is (I'm using 4.1.0)

PR #111 has been closed. @stefangordon's fix was looking good to me. 🙁

I've ended up with this ugly fix:
```javascript
const FixedCarousel = ({ children, initialSlideHeight, ...otherProps}) => {
  const slides = this.props.children.length 
    ? this.props.children 
    : <div style={{ height: initialSlideHeight }} />;

  return (
    <Carousel
      initialSlideHeight={initialSlideHeight}
      {...otherProps}
    >
      {slides}
    </Carousel>
  );
}

```

sarmeyer:
fixed in v4.3.10 after PR #412

PR #111, opened 2016-06-29
Initial slide height

stefangordon:
Per #105 - this appears to be an improvement in behavior.


stefangordon:
I'm not actually sure if this is better yet - there is still pretty big breaking behavior - closing until I can find a better solution.

PR #113, opened 2016-06-30
Add lazyLoad option and more examples in the demo

aij:
I tested with up to 10k slides. At that point it seems fine with devtools closed, but devtools adds a fair bit of lag, so I left it at 1k slides in the demo.


kenwheeler:
This looks great. Let me give this a solid run through first and then we'll get this merged.

aij:
Any chance this might make it into a release soon?

PR #114, opened 2016-07-01
Fix condition for goToSlide on props receive

obotor:
nextProps.slideIndex may sometimes be equal to 0, which results in falsy value.
nextProps.slideIndex should be tested as a number instead.


kenwheeler:
Well done

PR #130, opened 2016-07-23
Adding beforeSlide hook conditional sliding.

ericf89:
I needed a carousel that only transitions forward on certain conditions.  I made some changes to Nuka Carousel to check the return of the `beforeSlide` hook and only move forward when it's _not_ `false`. (`undefined` should be ok, hopefully ensuring backwards compatibility.) 

Not sure if this is something useful enough to be merged, but if so let me know and I can add some tests, and make sure it works for multi slides + different orientations.  I can't get the tests running, even on a clean master... so it'll take a little digging. 

```
PhantomJS 1.9.8 (Mac OS X 0.0.0) ERROR
  RangeError: Maximum call stack size exceeded.
  at ~/nuka-carousel/node_modules/chai/chai.js:3863
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,12 @@ Interval for autoplay iteration. Defaults to 3000.
 ####beforeSlide
 `React.PropTypes.func`
 
-Hook to be called before a slide is changed.
+Hook to be called before a slide is changed. Return `false` if you want to prevent the carousel from transitioning.
+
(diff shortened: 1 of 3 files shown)


ericf89:
Never had a chance to really test this extensively, and I think it breaks some other functionality.  Closin' time.

Issue #79, opened 2016-04-15
Allow dynamic slideToShow updating


Joneser:
When binding the slidesToShow attribute to a dynamic variable, it takes some time to update, I believe this is because the slidesToShow pop is not being updated in componentWillReceiveProps().

mankittens:
See PR #131

mariano-formidable:
The latest version of `nuka-carousel` (4.3.2) does handle `slidesToShow` prop updates within `componentWillReceiveProps`.  This issue should now be resolved.

PR #131, opened 2016-07-25
Make slidesToShow an updatable property

mankittens:
Solves two problems: (1) the component tries (and fails) to update itself when rerendered given the same props again, and (2) the component doesn't do a full update when given a new value for `slidesToShow`.

I get the feeling that there’s a better implementation.. but this should serve as a proof of concept.

In my use case, I'm changing the number of slidesToShow based on the screen width. This allows me to do this without things breaking.

Resolves Issue #79

### DIFF
--- a/lib/carousel.js
+++ b/lib/carousel.js
@@ -32,6 +32,8 @@ var _exenv = require('exenv');
 
 var _exenv2 = _interopRequireDefault(_exenv);
 
+var _lodash = require('lodash');
+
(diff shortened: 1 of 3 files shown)


Joneser:
Are we able to get this merged? Would be nice to get in and would certainly fix some issues I'm encountering

kenwheeler:
@mankittens can you resolve the conflicts and make sure this works?

mariano-formidable:
This issue seems to have been resolved in current versions of `nuka-carousel`.

PR #133, opened 2016-08-03
Fix slideHeight when heights vary and add lazyLoad option

aij:
Let me know if you'd like the slideHeight fix separated out. I just found it convenient to build upon the earlier changes to the demo app.

Or just pull #113 first. :)

### DIFF
--- a/demo/app.js
+++ b/demo/app.js
@@ -6,14 +6,14 @@ import ReactDom from 'react-dom';
 
 window.React = React;
 
-const App = React.createClass({
+const ControlledCarousel = React.createClass({
   mixins: [Carousel.ControllerMixin],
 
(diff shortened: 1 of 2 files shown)


kenwheeler:
Awesome work, would definitely prefer them separated, as i'd like to test independently

kenwheeler:
Speaking of tests, can you add one or two to check that this works right? not that I dont believe you that it works, i want to future proof myself from screwing this func up again.

uxtx:
This PR is fantastic! Thanks @aij! I read a few of the messages, and it looks like he's stepped away - but his work is solid I'd be happy to contribute some tests for these changes, as well as a few other untested functions.

PR #135, opened 2016-08-08
use relative positioning instead of absolute

fat:
I kept running into some weird issues with height not getting calculated quite right… And was wondering if you tried using position relative rather then absolute?

I noticed in an issue you mentioned that you used absolute for wrap around… but I switched up the math a bit and this seems to be working in demo (with wraparound and relative)

Unfortunately i couldn't get the tests running (so i may have totally broke everything :P)

If nothing else, thought it might be casually interesting to you.

thanks for the cool lib 🏄

### DIFF
--- a/demo/app.js
+++ b/demo/app.js
@@ -18,6 +18,7 @@ const App = React.createClass({
           ref="carousel"
           data={this.setCarouselData.bind(this, 'carousel')}
           slideIndex={this.state.slideIndex}
+          wrapAround={true}
           afterSlide={newSlideIndex => this.setState({ slideIndex: newSlideIndex })}>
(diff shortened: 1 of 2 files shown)


kenwheeler:
just seeing this now. this looks dope, I'll give it a spin.

kenwheeler:
@mdaxtman can you check this out?

mdaxtman:
This is good, thanks! @fat do you mind rebasing and resolving any conflicts, and running `gulp babel`? Otherwise I can pull locally and do it myself. Up to you. :-)

mdaxtman:
closed in favor of #161

Issue #139, opened 2016-08-30
For Vertical Carousels the Last Page of Slides Doesn't Fill Up the Available Space


cristianjd:
On horizontal carousels, when clicking to the last page of slides, the slides should be end to end filling up the available space. When the carousel is vertical, the same behavior is expected but is off a bit. See screenshot below:
![screen shot 2016-08-30 at 2 29 10 pm](https://cloud.githubusercontent.com/assets/4043331/18107791/443df50a-6ebe-11e6-9b6d-e4209b865ab7.jpg)

To reproduce this issue:
Number of Slides: 6
`slidesToScroll: "auto"`
`slidesToShow: 5`
`cellSpacing: 12`

PR #141, opened 2016-08-31
Fix slideToScroll='auto' bug

rawrmonstar:
Fixes bug where when `slideToScroll` prop was set to auto, the last
page of slides would have a gap after the last slide.

Addresses #139


kenwheeler:
Thanks!

PR #143, opened 2016-08-31
Fix slideToScroll='auto' bug including when the last slide is partially visible

rawrmonstar:
> Fixes bug where when slideToScroll prop was set to auto, the last page of slides would have a gap after the last slide.

Addresses #139 and includes fix which caused the reversion of original pull request #141 with #142.


rawrmonstar:
@kenwheeler can you have a look at this PR when you get a chance?

PR #146, opened 2016-09-20
Lazy load + Height + tests

uxtx:
adding a few unit tests for the already solid work @aij had done. Thanks again!

Included tests for
- lazyScroll
- heightMode

Also included some tests for
- vertical carousel
- wrapAround
- slidesToScroll

Also, i pulled out some duplicated code into some test helper utilities - hope that's ok.

### DIFF
--- a/demo/app.js
+++ b/demo/app.js
@@ -6,14 +6,14 @@ import ReactDom from 'react-dom';
 
 window.React = React;
 
-const App = React.createClass({
+const ControlledCarousel = React.createClass({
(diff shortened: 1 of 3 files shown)


angusfretwell:
So it looks like this has vanished from master (did someone force push an old branch 😛)? Keen to use the lazy loading functionality.

adamawang:
Seems like lazy load was included in the 2.1.0 and 2.1.1 releases, but then was rolled back to 2.0.4 that does not include lazy load. Will lazy load be put back into the next release?

PR #150, opened 2016-10-13
Updated karma-phantomjs-launcher

ar7n:
PhantomJS 1.9.8 craches in macOS Sierra, so it's imposible to run tests on latest macOS


ryan-roemer:
We've switched to Jest, so no more phantom!

PR #110, opened 2016-06-29
fixed issue where unable to reach last item(s) with slidesToShow > 1

awhitehouse104:

Issue #96, opened 2016-06-07
Can't scroll to the last items on the right when slidesToScroll={3}


gkrinc:
When I have a carousel with 9 slides and the following settings, I can't seem to scroll to the last 2 items on the right. I get the elastic effect and the highest currentSlide will go is 6. If I set slideIndex to 8 in the settings currentSlide is 8 and I can scroll to 7, then 6, but once I'm at 6 I can't go back to 7 or 8.

```
var settings = {
       slidesToShow: 3,
       slidesToScroll: 1,
       cellAlign: 'center',
       cellSpacing: 0,
       slideWidth: 1,
       easing: 'easeInOut',
       decorators: null,
       slideIndex: 2
};
```

If I change the slidesToShow to 1 I can get to all of the slides; currentSlide will go up to 8.

Here's the state when currentSlide is 0 and 6
![screen shot 2016-06-07 at 4 43 12 pm](https://cloud.githubusercontent.com/assets/2718382/15874064/80f7c1c8-2ccf-11e6-8e43-2b9a870b9853.png)
![screen shot 2016-06-07 at 4 43 50 pm](https://cloud.githubusercontent.com/assets/2718382/15874065/80f8a886-2ccf-11e6-9d72-cbe7cc9f5ad3.png)

Using v1.2.1

kenwheeler:
Can you try with master?

gkrinc:
Still not working on master (2.0.1) and I ended up with a few other visual bugs

gkrinc:
@kenwheeler, can you point me to a spot in the code where I could debug in my setup? I assume it has something to do with a bounds calculation of the last slide index or something along those lines.

mouhsinelonly:
+1 does not drag to last item

awhitehouse104:
This seems to be a problem with any number other than 1, not just 3

awhitehouse104:
Fixed this in [pr #110 0](https://github.com/kenwheeler/nuka-carousel/pull/110) for anyone interested

gkrinc:
Definitely related to slidesToShow > 1 and cellAlign = 'center' in my case.

ACPrice:
@kenwheeler Is this likely to be merged soon? I'm encountering this issue.

mdaxtman:
fixed with PR #159

gkrinc:
[...]

PR #159, opened 2016-11-23
Fix last slides not visible

mdaxtman:
@kenwheeler 

addresses #96, and is a more comprehensive fix than #110 will close both once this is merged.

nuka

- This PR takes into account what new prop `scrollMode` is, and doesn't show empty cells when finished scrolling if that prop is left blank or set to `remainder`.

- If `slidesToShow` and `slidesToScroll` are not mathematically divisible into each other, this PR ensures that the decorators function correctly to navigate to the last slide in the deck, even if the last slide index < `currentSlide` + n of `slidesToScroll`.

- `cellAlign` is taken into account so that whitespace is not shown if `scrollMode === "remainder"`

- Navigation through decorators as well as swiping are taken into account with this fix.

### DIFF
--- a/lib/carousel.js
+++ b/lib/carousel.js
@@ -852,7 +852,7 @@ var Carousel = _react2['default'].createClass({
       }
 
       if (index <= slidesAfter - 1) {
-        return slidesBefore < 0 ? (this.state.slideWidth + this.props.cellSpacing) * (this.state.slideCount + index) : 0;
+        return (this.state.slideWidth + this.props.cellSpacing) * (this.state.slideCount + index);
       }
     }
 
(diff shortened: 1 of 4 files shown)


mdaxtman:
revert a commit that was off of a PR that I pulled locally and forgot that I branched off of.

kenwheeler:
@mdaxtman we good to go?

gkrinc:
This helped a bit, but I still can't scroll to the last slide. I couldn't get to the last 2 items before, so this is a step in the right direction. 

These are my settings:

```
slidesToShow: 3,
slidesToScroll: 1,
cellAlign: 'center',
scrollMode: 'remainder',
```

Total slides: 9

The 8th slide centers and the 9th is visible on the right, but I can't scroll to the 9th slide.
Is

mdaxtman:
@gkrinc go ahead and set `scrollMode: 'page'` in your props. That's a new prop that just got merged in that should give you the behavior you're looking for. "page" should ensure that it will scroll the last slide to your 'cellAlign' position, and "remainder" should scroll until the last slide is visible, no matter what your alignment position is.

I'll set up the documentation for that.

PR #166, opened 2016-12-07
Fix currentSlide when number of children and slideIndex changes.

lauriejones:
The number of children in our carousel changes based on our app state. We also dynamically alter the initial slideIndex based on state.

This fixes an issue when the carousel updates from 3 children to a higher number, with a slideIndex greater than the previous number of children.

This uses nextProps for goToSlide in componentWillReceiveProps, whilst maintaining the previous behaviour of `goToSlide` using `this.props`.

Thanks for your consideration!


M-OC:
We're currently dealing with this exact issue.

Issue #129, opened 2016-07-21
afterSlide fires when slide does not change


BuffaloBuffalo:
The documentation for the `afterSlide` event states:

> afterSlide
> `React.PropTypes.func`
> 
> Hook to be called after a slide is changed.

However, it seems like `afterSlide` is called regardless of whether the slide has changed or not.  To reproduce, add a afterSlide callback like:

```
afterSlide={(slideIndex)=>{
  console.log('new slide index', slideIndex);
})}
```

Now, when if a slide is slightly dragged but the shown slide does not change, afterSlide will continue to print to the console.

The current workaround to this is writing a function like

```
afterSlide={(slideIndex)=>{
  if(this.state.slideIndex !== slideIndex){
      console.log('new slide index', slideIndex);
  }
})}
```

 However, this seems like it should be unnecessary given the phrasing of the documentation.

Jonwheeler:
I am running into this as well, the event appears to fire anytime the Carousel component is clicked.

PR #169, opened 2016-12-21
Don't fire goToSlide if slide was not changed

komachi:
It's also a better fix for #129


ashika01:
I’m closing this PR because of outdated code. We can always reopen it though! :) Please pull our recent version and create a new PR with our template if you think we still need this.
Thank you for contributing. If you need any help, please let us know.

PR #173, opened 2017-01-04
[Patch] Fix NPE when children are null or undefined

niki4810:
### Summary
- When the carousel children are null or undefined, added a null check condition within formatChildren function to prevent an NPE.
- Updated demo page to image urls

### Screenshots
nuka-npe-bug

//cc @kenwheeler


kenwheeler:
Can you rebase this against current master?

mariano-formidable:
Current versions of `nuka-carousel` simply use the `index` as the `key` and no longer check if a child component has an `id` prop, therefore this bug will no longer happen.

PR #183, opened 2017-01-13
Adding a flag to skip adding styles for image sliders

### LINKED ISSUE #182: Is that hardcoded style tag for slide images needed?
I'm not using images on my slides - nor the original decorators - and I found on the generated code a [weird style tag](https://github.com/FormidableLabs/nuka-carousel/blob/master/src/carousel.js#L818) that doesn't make much sense... It feels more like a CSS rule related to the samples?
I wonder if there could be an option to disable it? Or maybe the original code could verify if there's an image in between its direct children and add the style if needed?

`<style type="text/css">.slider-slide > img {width: 100%; display: block;}</style>`

igorsantos07:
This closes #182

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,6 +149,11 @@ Initial height of the slides in pixels.
 
 Initial width of the slides in pixels.
 
+####notImageSlider
+`React.PropTypes.bool`
+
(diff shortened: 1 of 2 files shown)


mariano-formidable:
This PR is too old to merge, but PR #378 was created to address the same issue in the current build.

Issue #168, opened 2016-12-14
draggable={ false } prevent slide content iteraction


lamartire:
Hello!
When I use carousel with not-image elements and set `draggable={ false } `, all attempts to interact with content are fails.
Example:
```
onClickButton = (e) => {
  console.log(e);
}

...

<Carousel>
  <div>
    <p>Slide 1</p>
    <button onClick={ this.onClickButton } value='Click me!' /> // Not clickable
  </div>
</Carousel>
```

melissa-hong:
Hi, 

I've run into this issue too and it looks like it's because this.clickSafe has a default of 'true' which prevents default on the event (in handleClick). When draggable is set to true, it dynamically sets clickSafe depending on whether the user is swiping so it behaves correctly (prevents default only when the user is in the middle of a drag/swipe):

```
handleClick: function handleClick(e) {
    if (this.clickSafe === true) {
      e.preventDefault();
      e.stopPropagation();

      if (e.nativeEvent) {
        e.nativeEvent.stopPropagation();
      }
    }
  },

  handleSwipe: function handleSwipe(e) {
    if (typeof this.touchObject.length !== 'undefined' && this.touchObject.length > 44) {
      this.clickSafe = true;
    } else {
      this.clickSafe = false;
    }
```

I think the solution is to default clickSafe to be false when it's defined. In the meantime a quick fix is to call e.stopPropagation() on all your click events so it doesn't bubble up to the carousel.

ritwik:
I believe the real issue is that clickSafe is used to protect against drags / swipes, but when returning the list of drag / touch functions, it doesn't account for the fact that no clicks will be possible without the handleSwipe function having been called. The tricky thing is that clickSafe needs to be true for either swiping or dragging (or both), but not when both are false.

I think a potential solution is to update the default to false, and then modify these two bits of code to switch it to true if either dragging / swiping is allowed:
```javascript
if (self.props.swiping === false) {
  return null;
} else {
  this.clickSafe = true;
}

...

if (self.props.dragging === false) {
  return null;
} else {
  this.clickSafe = true;
}

```

isadovskiy:
Guys, any update on the fix for this issue?

varenya:
Even anchors tags are not working because of this. any work around for that?

jflayhart:
Here's a PR: https://github.com/FormidableLabs/nuka-carousel/pull/423

sarmeyer:
fixed in #423

PR #190, opened 2017-02-01
Fix clickSafe preventing all clicks when dragging / swiping are both disabled.

ritwik:
This fixes issue #168.

When both swiping and dragging are disabled, clickSafe isn't needed yet it is set to true by default because the handleSwipe function is never called to unset it after a click. This means the click event is always blocked.

This change replaces that behaviour by defaulting clickSafe to false and enabling it if either dragging or swiping.

I've also added swipe and drag enable and disable buttons to the demo to aid testing (and showing the features of the project).

### DIFF
--- a/demo/app.js
+++ b/demo/app.js
@@ -9,7 +9,7 @@ window.React = React;
 const App = React.createClass({
   mixins: [Carousel.ControllerMixin],
 
-  getInitialState() { return { slideIndex: 0 }; },
+  getInitialState() { return { slideIndex: 0, dragging: true, swiping: true }; },
 
   render() {
     return (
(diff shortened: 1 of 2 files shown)


waynedunkley:
Bump for this fix

edorivai:
Also bumping for this fix. Also, wanted to mention that because of the `preventDefault`, this will eat clicks on basic `<a>` elements as well. So you won't be able to populate the carousel with links, or at least, users won't be able to follow the links.

edorivai:
A temporary fix for my problem (`<a>`'s not being clickable):

```js
<NukaCarousel>
	{ children.map((child, index) => <div key={index} onClick={e => e.stopPropagation()}>{child}</div>) }
</NukaCarousel>
```

It's really not pretty, but it get's the job done until this PR is merged.

ritwik:
@edorivai - Not sure this repo is maintained any more. My team has switched to using react-slick (https://www.npmjs.com/package/react-slick) for carousels like this and that's working fine for us.

ritwik:
@edorivai Sorry, spoke to soon - looks like it is still maintained & updated for fixes to react. But this issue has been open for a while now with no resolution (we moved off it so we didn't have to maintain a fork with fixes we wanted).

AVickory:
Same boat, given all the open PRs it seems like there's a pretty big pool of outside contributors who could be on-boarded into maintainer roles.

jflayhart:
Here's a PR: https://github.com/FormidableLabs/nuka-carousel/pull/423

itdsntwork:
I still get that error. Links are not working inside slides.

PR #193, opened 2017-02-08
Add Option to Remove Next / Prev Buttons

dtison:
The default Decorators place Next Prev buttons on the carousel.   These buttons could be removed by providing other Decorators.  This is intended an easier way for users to remove the Next Prev buttons only.

showNextPrev={false}


jasonleibowitz:
This is probably outdated now that v4 is out and decorators are deprecated? Should probably close it?

PR #200, opened 2017-02-23
Fix clickSafe preventing all clicks when dragging / swiping are both …

waynedunkley:
This is a duplicate of the fixes proposed by https://github.com/ritwik


AVickory:
bump for this fix.

Issue #203, opened 2017-03-15
Is there a way to disable an edge easing?


ivan-rudyi-personal:
@kenwheeler could you please help? Is this project alive ?

ivan-rudyi-personal:
@kenwheeler could you consider the fix for this issue https://github.com/FormidableLabs/nuka-carousel/pull/204 ?

auglean:
Have you tried setting the edge easing to linear? Not really sure what effect you want to obtain.

carlospaelinck:
The `edgeEasing` prop supports any D3 easing function names as strings.

PR #204, opened 2017-03-15
Is there a way to disable an edge easing? #203

ivan-rudyi-personal:


mariano-formidable:
The current versions of `nuka-carousel` no longer use `kw-react-tween-state` to tween animations so the coded solution in this PR is no longer applicable.

PR #211, opened 2017-04-15
Format README

LiHwsqh:
there should be a blank between '#' with the title word


alvaropinot:
It would be amazing to have this merged 🎉 😄 , seem pretty straightforward 😛

ryan-roemer:
Other folks also helped out and it's fixed! Thanks for helping too!

PR #206, opened 2017-03-22
Improve README readability

markmur:

PR #215, opened 2017-04-28
Fix markdown syntax

frankolson:
making the headers a bit more readable


frankolson:
is a duplicate of #206

PR #220, opened 2017-05-18
Fix drag to change slide

bPopovska:
When the carousel slide is clickable, dragging to change slide invoked the click handler as well.


mariano-formidable:
This PR is to old to merge, but issue #381 was created to resolve the issue in the current build.

PR #226, opened 2017-06-30
Fixes documentation formatting

vmattos:
The current doc seems to have some issues with markdown formatting


ryan-roemer:

PR #232, opened 2017-08-01
Fixes bad markdowns

ImanMh:


ryan-roemer:

PR #240, opened 2017-08-29
Fix slideCount being undefined

swaroopsm:
This uses `React.Children.count` for counting the children instead of `this.props.children.length` that was returning undefined when used in Decorators.


jasonleibowitz:
This is probably outdated now that v4 is out and decorators are deprecated. It should be closed?

PR #242, opened 2017-08-31
moving create-react-class to dependencies and fixpack

jonkwheeler:
- [ ] `create-react-class` is now in dependencies, not devDependencies

I also ran [fixpack](https://www.npmjs.com/package/fixpack) on the package to make sure it was ordered cleanly.


mariano-formidable:
The current version of `nuka-carousel` (4.3.2) does not have `create-react-class` as any type of dependency so this PR is no longer needed.

Issue #234, opened 2017-08-02
Doesn't create list item <li> element if one element


kosiakMD:
Carousel doesn't create list item <li> if only one element gave in 'children' array

foglerek:
This leads to margin issues, where if given a `cellSpacing`, the single item will have a negative margin of `-(cellSpacing/2)`.

foglerek:
Confirming that this problem still exists - it causes issues when providing a `cellSpacing` with only a single item present, as the item receives a negative margin that cuts it off.

PR #247, opened 2017-09-12
Fix Render <li> element even one child exists

kosiakMD:
Fixed Render <li> element even one child exists
Resolving Issue #234
https://github.com/FormidableLabs/nuka-carousel/issues/234

### DIFF
--- a/src/carousel.js
+++ b/src/carousel.js
@@ -166,10 +166,7 @@ const Carousel = createReactClass({
 
   render() {
     var self = this;
-    var children =
-      React.Children.count(this.props.children) > 1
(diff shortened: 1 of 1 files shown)


ChristiaanScheermeijer:
Any update on this?

RWOverdijk:
👍 

Weirder though, is that I don't understand why this conditional was here in the first place. I thought it might be so you can pass in the children you rendered yourself, but then why is the <ul /> always there? In any case, this is a needed fix for me.

@FormidableLabs what needs to be done to get this in?

RWOverdijk:
Bump.

Issue #82, opened 2016-04-19
translate3d with half pixel values causes blurry text in carousel


lauriejones:
At certain (odd) pixel screen widths the translate left position gets a half pixel value.

e.g. `transform: translate3d(-325.5px, 0px, 0px);`

This causes all text within the carousel to be blurry. Can this value be rounded to a whole number?

`this.getTweeningValue('left')` => `Math.round(this.getTweeningValue('left'))`

peteroid:
Not sure if this is related my issue #127

sarmeyer:
Hi @lauriejones - I'm going to close this issue as I don't believe it's applicable anymore with the updates that have been made to the carousel since 2016. If it is indeed still an issue for you, please reopen this issue and if possible, include steps to reproduce or a sandbox example. Thanks!

PR #251, opened 2017-09-15
Fix for #82: fractional values for translate3d cause blurriness

kartikluke:
Fractional values for translate3d cause a blurring effect in most browsers.


kartikluke:
@kenwheeler I've tested this fix out in production. It's a simple fix that improves the look a lot. Couldn't figure out why the blur was showing up till recently.

kartikluke:
Rebased this on master.

kartikluke:
@kenwheeler Could you review this?

wmertens:
Would it not be better to return a rounded integer from `getTweeningValue` instead?

kartikluke:
Looks like nuka-carousel is no longer using react-tween-state.

PR #253, opened 2017-09-20
Bugfix/after slide invoke

Robin-front:
afterSlide should always call after animateSlide completed

trace: ant-design/ant-design-mobile#1870


schnubor:
would love to see that merged!

mariano-formidable:
This PR is too old to merge, but we've created issue #376 so that we can implement a solution in our newer build.

PR #261, opened 2017-10-19
Update packages and update to webpack 3

daphnesmit:


ryan-roemer:
Thanks for the infrastructure help -- we're actually in the middle of a major overhaul to the code + deps and we've separately caught up to webpack4! (https://github.com/FormidableLabs/nuka-carousel/blob/master/package.json)

PR #272, opened 2017-12-05
Add ability to use translate over translate3d

Klaasvaak:
The webapp I'm working on was having scroll performance issues on Chrome for Android for the entire page, the swiping through carousel did work correct. When debugging this issue I noticed nuka-carousel component was causing this issue. When using `translate` instead of `translate3d` it was fixed. I think this is because the content of my slides where to heavy for `translate3d`

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,6 +186,11 @@ Animation duration.
 
 Enable touch swipe/dragging
 
+### use3d
+`React.PropTypes.bool`
+
+Enable to use CSS property translate3d instead of translate, this can change scroll performance. Defaults to `true`.
+
 #### vertical
 `React.PropTypes.bool`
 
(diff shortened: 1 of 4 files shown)


wmertens:
Hmm, you're changing the default here - shouldn't this be opt-in instead of opt-out? In the majority of cases `translate3d` is better, no?

Klaasvaak:
@wmertens translate3d is used as default, also in this PR.

wmertens:
Argh I missed the defaults true prop - sorry for the noise. LGTM!

wmertens:
@Klaasvaak did you close this on purpose?

Klaasvaak:
@wmertens yes, I don't see any progress for this PR. I also don't see the urgent of the PR any more since the new versions perform much better.

PR #279, opened 2018-01-24
Height features

ns4style:
Hello.
I find similar pull-request, but it's not be accepted.
Now i added heightMode again ( height will be equal a height of first slide or max height among their).
I added needRecalculateHeight props, that will be recalculate height of carousel, when one of children was updated.
I added some tests for these features and included this in README.


ryan-roemer:
Closing in favor of https://github.com/FormidableLabs/nuka-carousel/pull/291 which is merged and released -- it was based on your work so thanks!!!

Issue #282, opened 2018-02-08
Infrastructure: Update `es` build, modern webpack, yadda, yadda


ryan-roemer:
- [x] Remove `gulp`. Switch to npm workflows.
- [x] Remove + `.gitignore` `package-lock.json` and switch to `yarn`.
- [x] Update Travis to use Yarn.
- [x] Use an `.npmignore` whitelist to remove all the unnecessary cruft you can currently see in: https://unpkg.com/nuka-carousel@3.0.1/
- [x] Bump all relevant dependencies.
- [x] Update babel stuff
- [x] Update webpack versions
- [x] Export an `es/` directory
- [x] Remove `lib` from git source. Fixes #224
- [x] Replace phantom in Karma with headless chrome. (See Radium)

ryan-roemer:
Headless chrome is only a work item if we're still doing PhantomJS.

Jest just uses jsdom so not needed for any tests using Jest.

We could separately do functional tests with headless chrome, but I'd prefer we just ticket out and scope new work like that to properly plan.

carlospaelinck:
Yeah. That sounds good.

ryan-roemer:
Fantastic work @carlospaelinck !!!

I've moved the remaining `npm version` work (which is one of our final things for v4 release) to https://github.com/FormidableLabs/nuka-carousel/issues/295 with specific instructions.

PR #288, opened 2018-03-01
Upgrade build processes and dependencies

carlospaelinck:
* Upgrade to Webpack 4 and Babel 7
* Use `builder` and `babel-cli` to make ES and CommonJS outputs
* Remove `lib` from source
* Update development demo build and distribution Webpack build
* Add support for Formidable ESLint React ES6 config
* Refactor string-based `ref`s to function-based ones
* Update the Nuka source to conform to ESLint rules
* Add support for Prettier
* Add `yarn.lock` which makes this PR look much bigger than it is
* Bump version number to 4.0.0

Partially addresses #282 
A refactoring of tests will be in a subsequent PR

cc @ryan-roemer @kenwheeler


.eslintrc
```diff
@@ -1,75 +1,27 @@
...
-        "eol-last" : 1,
-
-        "comma-spacing": [1, {"before": false, "after": true}],
-        "indent": [1, 2],
-        "quotes": [1, "single"]
+  "extends": ["formidable/configurations/es6-react", "prettier"],
+  "env": {
+    "node": true,
+    "browser": true,
+    "mocha": true
```
ryan-roemer: Maybe remove this and add a `test/.eslintrc` that extends this one plus adds it there?

Or, if bloating this PR, just separately ticket for later? (Not a huge priority)

package.json
```diff
@@ -1,31 +1,40 @@
...
-    "eslint-loader": "^1.0.0",
+    "eslint-config-formidable": "^3.0.0",
+    "eslint-config-prettier": "^2.9.0",
+    "eslint-loader": "^2.0.0",
     "eslint-plugin-babel": "^4.1.2",
+    "eslint-plugin-filenames": "^1.2.0",
+    "eslint-plugin-import": "^2.9.0",
+    "eslint-plugin-prettier": "^2.6.0",
     "eslint-plugin-react": "^7.4.0",
     "gulp": "^3.9.0",
```
ryan-roemer: Can we remove all the gulp deps and the`Gulpfile`? Or is that part of the tests?

carlospaelinck: Yeah, tests don't work at all anymore. We're going to move to jest. So I can just remove that stuff for now.

.eslintrc
carlospaelinck: I am going to remove all test stuff and start fresh. But keep the spec files to ensure we're maintaining coverage when we move to jest.

carlospaelinck:
@ryan-roemer I removed the gulp and Karma stuff and added the npmignore.

When I run `npm pack` it does look correct as I see this:

screen shot 2018-03-01 at 4 05 27 pm

.eslintrc
ryan-roemer: Sounds good!

ryan-roemer:
Notes on the CI failures:

- If you're using `webpack4` you need node 6.11.something or above. So drop node4 from the matrix. (Maybe add node9 for future goodness)
- The other node versions are failing for invoking gulp. You probably need to scour all text files for incidences of gulp to remove :)

carlospaelinck:
That was my next thing to look at before I merge. Thanks!

carlospaelinck:
That was also my reasoning for the version bump because I think the goal is to also move away from mixins and `createReactClass` too in addition to updating tooling.

Issue #292, opened 2018-03-07
Refactor Nuka to use `React.Component` and modern React patterns


carlospaelinck:
- [x] Replace `react-tween-state` and mixin logic
- [x] Refactor implementation to use `React.Component` classes instead of `createReactClass`

PR #293, opened 2018-03-09
Refactor Nuka to use modern React

carlospaelinck:
- [x] Replace `react-tween-state` with `react-move`
- [x] Replace `createReactClass` with `React.Component`
- [x] Add test cases for `wrapAround` mode

Addresses #292 

cc @kenwheeler @ryan-roemer


carlospaelinck:
@ryan-roemer Mixins are a thing of the past with this PR.

ryan-roemer:
Looks like we still have one reference to `mixin` remaining in docs? https://github.com/FormidableLabs/nuka-carousel/blob/master/README.md#data

carlospaelinck:
Good catch.

Issue #270, opened 2017-11-27
reference internal carousel controls by separate props instead of decorators array


mvanroon:
What is the idea behind using:

```
<Carousel
    decorators={[PrevButton, NextButton, Dots]}
>
```

instead of for example:

```
<Carousel
    prevSlide: PrevButton,
    nextSlide: NextButton,
    dots: Dots
>
```

If you agree that the latter might work better, I'll make a PR. I won't if there's good reason you went for the decorators array.

Cheers

carlospaelinck:
There are 8 positions for decorators which we are going to rename controls. There will be render props for each of the 8 positions for version 4.0.

mackermans:
Hi @carlospaelinck, not sure if I missed a changelog somewhere, however, I would definitely appreciate knowing what breaking changes were introduced and how I would have to upgrade from v3 to v4 (and I'm sure other people too, that include your package in their projects). It took me a little while to figure out what changed, but I managed to upgrade in the end by finding this issue and changing the decorators to render props.

Anyway, just a suggestion coming from a slightly painful experience upgrading 👍 

Thanks!

carlospaelinck:
@mackermans Updated the release entry with changes.

Issue #285, opened 2018-02-17
Feature: adaptive height


ajoshi0:
## Updated ticket

- [x] ~~Get adaptive height in `3.x` branch~~ (fixed in #291)
- [x] Get adaptive height port of #291 into `master` branch after #293 lands.

## Arunesh's Original Description

please check these pr's and merge them which one is appropriate

https://github.com/FormidableLabs/nuka-carousel/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aopen+height

Currently slick supports adaptive height but NUKA does not.

cc @kenwheeler @ryan-roemer

johnking:
I just started using the tool, and spent a few hours on the the height issue.  Good to know it is on the milestone.

Nehle:
@ryan-roemer -- trying to build something off of the current PRs that are out there, but it seems the last PR merged removed the testing and broke the demo. Any ETA on when those will be working again so I can proceed?

cc @ajoshi0 @kenwheeler @alexlande

ryan-roemer:
@Nehle -- @carlospaelinck is actively working on getting the entire infrastructure in modern shape and I think https://github.com/FormidableLabs/nuka-carousel/pull/289 gets us back in "open for business" shape.

Assuming that's the case (is that right Carlos?), we should be ready to pick one of these approaches and have someone open a new PR to fix + add regression tests (using Jest now)

Nehle:
Excellent, I'll rebase my local branch on that. Thanks

carlospaelinck:
Yeah. This PR will finish the tooling. We will also be migrating the code base away from a Mixin-based approach to a modern style as well. So things will be in flux for a bit.

ajoshi0:
@carlospaelinck can we try to keep prop names and functionality as close as possible to current one if possible

carlospaelinck:
Yes, the goal is an internal refactor.

Nehle:
Is there any chance we can get a `3.x` branch to push the height feature to while we're waiting for the upgrade to be completed?

ryan-roemer:
@kenwheeler @carlospaelinck -- How far off is https://github.com/FormidableLabs/nuka-carousel/pull/289 from merging? I _think_ once we land that `master` is back open for business right?

carlospaelinck:
[...]

PR #299, opened 2018-03-12
Convert decorators to render props

carlospaelinck:
- [x] Remove decorators and decorators array
- [x] Add render props for each of the 8 positions for controls
- [x] Update demo

cc @kenwheeler @ryan-roemer 

Addresses #270


src/default-controls.js
```diff
@@ -0,0 +1,106 @@
...
+  background: 'rgba(0,0,0,0.4)',
+  color: 'white',
+  padding: 10,
+  outline: 0,
+  opacity: disabled ? 0.3 : 1,
+  cursor: 'pointer'
+});
+
+export class PreviousButton extends React.Component {
+  handleClick = e => {
```
ryan-roemer: GLOBAL: Is there any chance we can get rid of all member-as-arrow functions? The stuff babel produces is (1) not in the inheritance chain, (2) can't be introspected (because not on prototype), and (3) is stage-1.

Actually, could we maybe remove all of:

```
    "@babel/plugin-proposal-class-properties",
    "@babel/plugin-proposal-object-rest-spread",
    "@babel/plugin-proposal-optional-chaining"
```

and stick with env + react?

asimqt:
@ryan-roemer @carlospaelinck Hey, I need to use these render props instead of decorators. When are you going to publish V4.0.0?

ryan-roemer:
@asimqt -- We're blocked on the port of adaptive height from `v3.x` branch into `master` in #285 

(...or we have to do something awkward like remove adaptive height in `v4.x` and wait for someone to add it back in later...)

/cc @Nehle @ajoshi0 @tchon

asimqt:
@ryan-roemer What do you mean by adaptive height?  `heightMode` param with value `current` will do it, right? Or are we removing that also?

ryan-roemer:
@asimqt Published version `v3.x` and git branch `3.x` has an adaptive height feature that is *not* in `master` (aka soon to be `v4.x`) that needs to be ported.

See #285 and the [`v4.0` milestone](https://github.com/FormidableLabs/nuka-carousel/milestones/4.0)

PR #301, opened 2018-03-14
Fix currentSlide when deleting slides

Zephir77167:
When deleting slides, currentSlide could end up out of bounds. Now fixed

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -111,6 +111,9 @@ export default class Carousel extends React.Component {
     this.setState({
       slideCount: nextProps.children.length
     });
+    if (nextProps.children.length <= this.state.currentSlide) {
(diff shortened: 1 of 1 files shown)


Zephir77167:
Hey, I have little to no experience from testing components, could you please guide me through it?

Basically, there is an initial state where we have
```
<Carousel>
  <Child1 />
  <Child2 />
</Carousel>
```

Then we swipe to Child2, and update the state so that it is now
```
<Carousel>
  <Child1 />
</Carousel>
```

And before the fix, the carousel would still have `this.state.currentSlide === 1`, and would therefore be out of bound.

ryan-roemer:
Thanks for identifying the issue! We separately fixed this in https://github.com/FormidableLabs/nuka-carousel/pull/303

Issue #315, opened 2018-03-26
Wrap around mode incorrectly aligns the last slide when Carousel is internally controlled


carlospaelinck:
When using wrap-around mode and not using external control, the last slide does not wrap around correctly when navigated from the first.

michaek:
@carlospaelinck I'd add that the first slide does not display correctly when navigated from the last.

PR #320, opened 2018-03-29
Ensure left is getting set correctly on wrap around

carlospaelinck:
- [x] We only want to do wrap around offsets when we are wrapping around or dragging

cc @ryan-roemer @chrisbolin @kenwheeler 

addresses #315

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -829,7 +829,10 @@ export default class Carousel extends React.Component {
     const end =
       (this.state.slideWidth + this.props.cellSpacing) * slidesToShow * -1;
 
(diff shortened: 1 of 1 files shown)


lotap:
This broke my styling. I have a wraparound carousel in which the next and previous slides are supposed to be partially visible to give users a visual cue that they can click through in both directions.

After this change, no preview for the previous slide shows up at index 0 (when the page loads). The next slide preview also does not appear after reaching the last slide.

Can we put this change behind a prop like wraparoundPersistentPositioning that defaults to false but could be set to true to provide the old behavior?

Issue #318, opened 2018-03-28
Memory usage and garbage collection


michaek:
It seems that nuka-carousel creates a fair volume of new objects in memory as it's changing between slides. When `autoplay` and `wrapAround` are both true, that means that its consistently making objects and releasing them to garbage collection, leading to a heap size that looks like this over time:

![image](https://user-images.githubusercontent.com/155459/38050730-fe6ec616-3299-11e8-9ff3-6002a0437902.png)

That may not be a big deal, as references to these objects aren't kept, so they are getting collected - so long as the garbage collector is not otherwise indisposed. Where I'm running into a problem is leaving a tab containing a carousel open (in Chrome) and then returning to it after some time. The tab freezes entirely for some time before returning to normal. As far as I can tell, there's no issue with runtime code (JS CPU profiling looks normal), and what we're seeing is a GC pause as V8 cleans up after objects created in the background. Unfortunately, the frozen state also blocks the Dev Tools profiler from starting, so I don't exactly have a smoking gun...

However, it's pretty easy to reproduce this in a few minutes on the nuka-carousel demo:

```
    this.state = { slideIndex: 0, length: 6, wrapAround: true, autoplay: true, autoplayInterval: 100 };
```
> I made the autoplayInterval really short to make it consume memory faster. It's kind of unpleasant (and maybe unsafe for some) to look at.

and
```
          wrapAround={this.state.wrapAround}
          autoplay={this.state.autoplay}
          autoplayInterval={this.state.autoplayInterval}
```

Once that's running, put the tab in the background for a few minutes by switching to another tab and then return to it. The pause should be noticeable.

If it does turn out to be related to work being done while in the background, perhaps a first step could be as simple as listening for visibility with the [Page Visibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API)?

michaek:
Using the Page Visibility API to set `autoplay` on and off from the calling component resolved this issue for my implementation. I think it makes sense to build that into nuka-carousel, as leaving `autoplay: true` leads to this difficult-to-troubleshoot bug that's pretty severe from a user experience perspective.

michaek:
I think #321 is a great change, and it should dramatically cut down on the scale of this issue. The carousel should ideally still stop doing work when the `document.hidden === true` though, right?

alexlande:
Your example locked up my browser, so yeah, verified! Will look into this.

michaek:
[...]

PR #321, opened 2018-03-30
Cut down object creation during transitions

sghall:
Hey,  author of react-move here.  Just happen to see https://github.com/FormidableLabs/nuka-carousel/issues/318.  I think this might help.  If I understand correctly, these objects just need to be created once per render of the carousel.  Inside the children function of Animate these objects are being created on each frame of the animation.  I did run the tests on this and everything passed.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1025,6 +1025,10 @@ export default class Carousel extends React.Component {
         ? 0
         : this.props.speed;
 
+    const frameStyles = this.getFrameStyles()
+    const touchEvents = this.getTouchEvents()
+    const mouseEvents = this.getMouseEvents()
+
     return (
       <div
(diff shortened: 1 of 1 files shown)


carlospaelinck:
@sghall Thank you, this is a great optimization.

ryan-roemer:
Fantastic work @sghall ! Thanks!

_Side note_ -- Our lint isn't working due to the `prettier` extends in eslintrc disabling a whole ton of things without us having something actually "fix" per prettier. will fix in follow on workl

sghall:
Right on.  Glad to help out!

ryan-roemer:
Released in `nuka-carousel@4.0.3`

PR #325, opened 2018-04-04
Handle swipe properly when carousel has one slide and disable the arrows

mdimovska:
* Extend the demo to include option for showing only one slide
* Disable the arrows when there is only one slide
* When swiping and there is only one slide, apply easing, instead of going to the next/previous slide

* Move getStyleTagStyles from carousel to the demo app (the user should be responsible to apply the proper styles of the slides)

### DIFF
--- a/demo/app.js
+++ b/demo/app.js
@@ -10,9 +10,13 @@ class App extends React.Component {
     this.state = { slideIndex: 0, length: 6, wrapAround: false };
   }
 
+  getStyleTagStyles() {
+    return '.slider-list img {width: 100%; display: block;}';
+  }
+
   render() {
     return (
-      <div style={{ width: '50%', margin: 'auto' }}>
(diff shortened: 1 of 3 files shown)


mariano-formidable:
@mdimovska - Thanks for the work in this PR.  It looks good, but we'll need the following updates done before we can accept it:

- Remove modifications to the demo app (the demo app already has a bunch of buttons toggling on/off features, we're trying to limit when we add in more).
- Return `getStyleTagStyles` back into the carousel as it was.  We are currently working on a feature (based off PR #183) to disable generation of that style tag if the user does not need it (but, by default, we've found many people report bugs that the carousel isn't working when that style is missing).
- Finally, could you add in some tests to `test\specs\carousel.test.js` to validate that next/previous arrows don't exist on a single slide?

Thanks again for your work.  Your involvement in our nuka-community is much appreciated!

PR #328, opened 2018-04-16
[PATCH] add onResize callback

Sebastiansc:
## Summary

* Add callback to allow components to adjust `lazyLoadIndex` on resize without implementing a resize listener


src/carousel.js
```diff
@@ -726,6 +728,7 @@ const Carousel = createReactClass({
   },
 
   onResize() {
+    this.props.onResize();
```
Nehle: Shouldn't this fire _after_ the resize is done?

Nehle:
This should be a `MINOR` update @Sebastiansc

src/carousel.js
ajoshi0: +1

klammm: +1

ryan-roemer:
@ajoshi0 @Sebastiansc -- Can you please target `FormidableLabs:master` and not `FormidableLabs:3.x`. We're not doing any future work on that branch.

Also, we've got easy-to-write jest tests now, so would appreciate a simple test to make sure everything gets called the right way. Thanks!

/cc @carlospaelinck

ajoshi0:
@ryan-roemer We can do another pr for master for same but we do need this feature on 3.x as we have lot of code dependent on 3.x. Plan is to migrate to 4/master soon but for an internal launch we have to use 3.x for now.

cc @Sebastiansc please open another pr for master and write a small jest unit test please

ryan-roemer:
@ajoshi0 -- We put a **lot** of effort into v4 and maintaining a double build (especially the `3.x` branch that commits `lib` to source, etc.) is something that is too costly for us to do. And, most relevantly, we don't have any Formidable who has free time right now to invest that kind of time.

Is there any specific blocker for migrating to `v4` that you see? We also took pains to make the changes as minimal as possible to the actual API surface.

ajoshi0:
@ryan-roemer I know at this point migrating it internally for next 2 weeks is not possible. Plan is to migrate to v4 by end of month it impacts many things on our side like ADA issues, current carousel decorators with v4 are not working that has to be fixed. Please if we can get this in would be great on 3.x branch.

ryan-roemer:
@ajoshi0 -- You can always take a git-based dependency (since stuff is committed to source in `v3`) and go off that as an interim measure off a fork. Then after your 2 weeks, you can hopefully migrate to v4.

(I wish you had alerted us to your timeline and needs earlier as we could have had more time and planning to guide and help after all your work a month ago, after which we abandoned `v3`, which we were only doing as a one-time thing to help you then).

ryan-roemer:
Oh, and chatted with Carlos -- decorators were the only API change, but as a heads up, we don't intend to support them further. The approach is not as good as the new API and decorators as a general concept are not preferred in our ecosystem now.

ajoshi0:
@ryan-roemer Sure no issues so when we migrate to v4 we will make all the fixes on our end. 

But please if you can get this in today on 3.x branch that would be really great.

ryan-roemer:
@ajoshi0 @Sebastiansc  -- Today is going to be tight, even as a "for old time's sake" pulling folks off stuff (aka me and carlos) type thing.

In chatting, Carlos noted that `setDimensions` does `setState` stuff -- is all of that **done** by the time `props.onResize()` is called?

Sebastiansc:
@ryan-roemer I did considered adding a callback to setState. However they're independent of each other as they both will trigger a re-render, whichever goes first won't make any difference. If you're more comfortable with it that way I can set it as a callback to `setState`

ryan-roemer:
@Sebastiansc -- Ok.

test/specs/carousel.spec.js
```diff
@@ -518,6 +518,20 @@ describe('Carousel', function() {
...
+      component = ReactDOM.render(
+        React.createElement(carousel, null,
+          React.createElement('p', null, 'Slide 1'),
+          React.createElement('p', null, 'Slide 2'),
+          React.createElement('p', null, 'Slide 3')
+        ),
+        container
+      );
+      const spy = sinon.spy(component, 'onResize');
+      component.onResize();
```
ryan-roemer: This test isn't testing anything. You've _spied_ `component.onResize`, then **directly called it**.

You need to spy on `props.onResize` and call `component.onResize` at a minimum.

Sebastiansc:
@ryan-roemer updated to call prop as `setState` callback

src/carousel.js
```diff
@@ -866,6 +868,7 @@ const Carousel = createReactClass({
         top: props.vertical ? this.getTargetLeft() : 0,
       },
       function() {
+        self.props.onResize();
```
ryan-roemer: Do we need to call `onResize` on every `setDimensions` call?

I was imagining a more honed refactoring of:

```js
setDimensions(stateCb) {
  // stuff
  this.setState(STUFF, stateCb || function () {});
}

onResize() {
  // stuff
  this.setDimensions(this.props.onResize);
}
```

Sebastiansc: Yes you're right I didn't realize this was being called from other places. Updating

Sebastiansc:
@carlospaelinck adding tests

Sebastiansc:
@carlospaelinck @ryan-roemer please re-review

ryan-roemer:
Released in https://unpkg.com/nuka-carousel@3.2.0/

PR #331, opened 2018-04-19
Disable autoplay on page visibility hidden

### LINKED ISSUE #318: Memory usage and garbage collection

alexlande:
This pauses autoplay when the tab running the carousel isn't in the foreground. This prevents a wacky issue spotted by @michaek in which a carousel running in the background accrues memory usage without garbage collection, which can cause your browser to lock up entirely when the page moves back to the foreground.

Resolves #318

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -81,6 +81,7 @@ export default class Carousel extends React.Component {
     this.getTargetLeft = this.getTargetLeft.bind(this);
     this.onResize = this.onResize.bind(this);
     this.onReadyStateChange = this.onReadyStateChange.bind(this);
+    this.onVisibilityChange = this.onVisibilityChange.bind(this);
     this.setInitialDimensions = this.setInitialDimensions.bind(this);
     this.setDimensions = this.setDimensions.bind(this);
     this.setLeft = this.setLeft.bind(this);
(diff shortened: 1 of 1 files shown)


kenwheeler:
👍

kumarharsh:
What's the expected release date for this fix? Just discovered this issue, and on low powered systems, it can cause the page to crash very easily.

kumarharsh:
I think #321 was a great change which would prevent this issue from coming up so frequently. This PR pauses autoplay when page gets hidden - but it does leave out the case when a user <kbd>Alt</kbd>+<kbd>Tab</kbd>s to a different window. Page Visibility API notably doesn't handle that case. It can be implemented by adding `window.addEventListener('focus')` and `window.addEventListener('blur')`.

Would adding support for that use case in this project be something you'd consider, or should that be left in the userland?

ryan-roemer:
Ack, we missed the publish!

Issue #317, opened 2018-03-27
How to make all slides the same height as the tallest slide?


misstricky:
Hello! I'm a long-time user of slick-carousel and then the react port of slick-carousel. In nuka-carousel, is there a way to make all slides the same height as the tallest slide? I have `heightMode` set to `max` and `shouldRecalculateHeight` on as well. While the tallest slide does fully appear (nothing is cut off), the other slides are not that same height (they are their shorter heights). I've tried a large amount of CSS on the slides/wrappers to make all the slides stretch to all be the same height, but any CSS that succeeds in doing that, on refresh, causes no slides to appear (slides have a height of 0).

Any javascript I've tried to get the `slideHeight` manually on each slide results in it working for most slides but not the first slide which stays its original shorter height.

I'm sure I'm missing something super obvious, but is there a way to make all slides the same height as the tallest slide? Does anyone happen to have some working example code where they're able to do this?

Thanks!

alexlande:
Just checked this out and confirmed the issue. I worked around it in my test by adding `min-height: 100%` to `.slider-slide`. I think we might just be able to add that as a default style, but I'll do some testing to make sure. In the meantime, hopefully you can do the same in your project 👍

dfagan2:
I think this needs reopened, since the min-height was reverted:
https://github.com/FormidableLabs/nuka-carousel/pull/398

I'm seeing the issue @misstricky described above

PolGuixe:
I am experiencing the same. I believe it should be reopen.

sarmeyer:
If you believe it should be reopened, please respond with a minimal code example of how you're using Nuka and the issue you're seeing. A codesandbox makes it very easy for me to replicate the issue and figure out a solution. Thanks!

PR #334, opened 2018-04-23
Add min-height: 100% to slides

alexlande:
Resolves https://github.com/FormidableLabs/nuka-carousel/issues/317

This ensures that slides are at least as tall as their container. Right now, if you use `heightMode='max'` to ensure that slides are all as tall as the tallest slide, the slide container is sized appropriately but individual slides aren't. Applying a `min-height` lets you style slides more easily within the container, so you can do things like this:

<details>
<summary>Expand/Collapse</summary>

slides

</details>


ryan-roemer:

PR #341, opened 2018-05-04
update this.controlMap when component updates 

### LINKED ISSUE #339: Nuka doesn't react to render controls prop change
Hi,

I'm trying to upgrade to v4, and I have an issue with the new control render props. They work fine when set the first time, but upon changing their value (in my case I'm changing the previous/next control depending on when you are on mobile or desktop), the carousel doesn't rerender the controls.

I can see the renderCenterCenterControls and renderBottomCenterControls props change when I resize my browser window, but it's not reflected in the carousel. If I do a full page reload, I can see the correct controls being displayed. Even calling react forceUpdate after a resize doesn't fix it.

It worked well in v3 with decorators prop.

Thanks.

zhaostan:
This fixes #339 where render props don't update after the component updates

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -134,6 +134,20 @@ export default class Carousel extends React.Component {
     }
   }
 
+  componentDidUpdate() {
+    this.controlsMap = [
+      { func: this.props.renderTopLeftControls, key: 'TopLeft' },
+      { func: this.props.renderTopCenterControls, key: 'TopCenter' },
+      { func: this.props.renderTopRightControls, key: 'TopRight' },
+      { func: this.props.renderCenterLeftControls, key: 'CenterLeft' },
+      { func: this.props.renderCenterCenterControls, key: 'CenterCenter' },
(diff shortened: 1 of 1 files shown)


moimael:
LGTM ?

src/index.js
```diff
@@ -134,6 +134,20 @@ export default class Carousel extends React.Component {
     }
   }
 
+  componentDidUpdate() {
+    this.controlsMap = [
```
ryan-roemer: Can we collapse this to maybe a helper function since it's copy and paste from elsewhere in this file?

Also, can we get a regression unit test when you get a moment? Thanks!

@mhink -- If you've got time, can you chat with @zhaostan and maybe pick this up if he doesn't have time? Thanks!

mariano-formidable:
@ryan-roemer It looks like PR #354 also addresses this issue, but has tests and has a solution that doesn't duplicate the `controlMap` array.

mariano-formidable:
Going to go ahead and close this PR since PR #354 addresses the same issue and has tests already associated with it! 😄

PR #344, opened 2018-05-09
Sometimes when the children are dynamic slideCount is not calculated correctly when entering in componentWillReceiveProps

andressspinetti:
Eg:

```
<Carousel ...>
  <div className="elem">...</div>
 {elements.map(elem => (<div key={elem.id} className="elem">{elem.name}</div>))}
</Carousel>```


ryan-roemer:
@andressspinetti -- Thanks for identifying this. Can you add a regression test as well?

andressspinetti:
@ryan-roemer Sure

PR #346, opened 2018-05-11
Allow Clicking on Elements in Carousel when Dragging is Disabled

jasonleibowitz:
#### What does this PR do? 
Adds fix that enables clicking on elements in a Carousel when dragging is disabled. This fix addresses an old issue (#168) from 2016 that was resolved in an old PR (#190) from 2017, but was never merged in. 

The library has since been updated causing that old PR to have breaking changes. This PR is an updated version of that fix for Nuka Carousel v4.2.1

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -36,7 +36,7 @@ export default class Carousel extends React.Component {
     super(...arguments);
 
     this.displayName = 'Carousel';
-    this.clickSafe = true;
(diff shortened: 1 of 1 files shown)


jflayhart:
Here's another PR: https://github.com/FormidableLabs/nuka-carousel/pull/423

PR #354, opened 2018-05-30
Update controls on render*Controls props change

mdimovska:
**Changes:**
* Update controls on render*Controls props change
* Add a unit test

Fixes https://github.com/FormidableLabs/nuka-carousel/issues/339.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -31,23 +31,24 @@ const removeEvent = function(elem, type, eventHandle) {
   }
 };
 
+const controlsMap = [
+  { funcName: 'renderTopLeftControls', key: 'TopLeft' },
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -31,23 +31,24 @@ const removeEvent = function(elem, type, eventHandle) {
   }
 };
 
+const controlsMap = [
+  { funcName: 'renderTopLeftControls', key: 'TopLeft' },
```
mariano-formidable: Couldn't `controlsMap` remain within the constructor of `Carousel`?  It can still be defined as you have it with `funcName` assigned to strings instead of `func` assigned to `this.props` values, but since it's so specifically needed for `Carousel` I don't think it needs to be a global `const`.

PR #355, opened 2018-06-04
Prevent setState after component Unmounts

stanley92:
Hi @kenwheeler ,

I am currently facing an issue where nuka carousel continues to setState in the function goToSlide after the component unmount. Attached is a screenshot of what happens.

setstate

As such, I did a fork and added a few checks in the function to check if component did unmount. If it does, it will not proceed with the setState. Hope you will be able to look at it and merge into master. Feel free to let me know if you have comments or advise to do it better. 

Thanks for your time!

Best Regards
Stanley Ong

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -436,7 +436,9 @@ export default class Carousel extends React.Component {
   // Action Methods
 
   goToSlide(index) {
-    this.setState({ easing: easing[this.props.easing] });
+    if (!this.mounted) return;
+
+    if (this.mounted) this.setState({ easing: easing[this.props.easing] });
 
     if (index >= React.Children.count(this.props.children) || index < 0) {
       if (!this.props.wrapAround) {
--- a/src/index.js
+++ b/src/index.js
@@ -464,14 +464,16 @@ export default class Carousel extends React.Component {
           }),
           () =>
             setTimeout(() => {
-              this.setState(
-                { isWrappingAround: false, resetWrapAroundPosition: true },
-                () => {
-                  this.setState({ resetWrapAroundPosition: false });
-                  this.props.afterSlide(0);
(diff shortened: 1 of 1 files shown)


ryan-roemer:
@stanley92 -- Thanks for this PR! Can you:

1. Fix lint
2. Add a regression test that exposes the issue and would fail on `master`

?

Thanks!

mariano-formidable:
@kenwheeler @ryan-roemer I can verify that this is a legitimate bug in nuka.  `goToSlide` will call one of two `setTimeout` functions when it performs a `wrapAround` either from last -> first or first -> last.  If the component gets unmounted before the `setTimeouts` return, then the calls to `setState` within the `setTimeout` callbacks trigger the error.

This PR implements a solution that checks if the component is mounted before calling `setState`.  That fixes the issue, but it falls into the area of an antipattern (https://reactjs.org/blog/2015/12/16/ismounted-antipattern.html).  I think a more durable solution would be either to:

1.)  Clear any active `setTimeout` functions in the `componentWillUnmount` method.

-- OR --

2.) Remove the use of `setTimeout` functions from `goToSlide`.  Perhaps there is another way to perform that required `wrapAround` logic without requiring them.  This way the component can be unmounted at any time with no problems.

mariano-formidable:
@stanley92 - Thanks for catching this bug.  We're going to close this PR, but can you create a new PR with failing regression tests exposing the bug in `test/specs/carousel.test.js`.

While your solution definitely works, we think the problem lies deeper in the way we're handling transitions between slides (particularly when they wrap around).  Ideally, there shouldn't have to be any `mounted` checks within the code.  Because of the complexity, we'll take on the task to fix it, but a PR with failing tests would help us out.  Thanks again!

PR #369, opened 2018-07-05
Bug/first last partial wrap around

### LINKED ISSUE #194: When displaying more than 1 slide wrapAround doesn't function as one would expect(Re: infinite looping/head to tail)
An infinite wrap around should place the 1st slide at the end of the last slide.. creating an infinite loop. wrapAround when displaying multiple slides should be head to tail.

currently the way the implementation of wrapAround works is if I have 8 total slides, 6 visible and slide 1 slide at a time, it doesn't wrap until the last slide has passed creating a huge area of nothing

broken_head_to_tail

### LINKED ISSUE #324: First and last slide are not displayed when wrapAround is set and slidesToShow is a decimal number
Having the following options set:
```
cellAlign='center'
slidesToShow={1.25}
wrapAround
```
the last slide is not shown when the 1st slide is active:
image

Same applies for the first slide (it's not shown when the last slide is active):
image

The other slides are displayed properly:
image

**How to reproduce the issue:**
Set the following options in the Carousel in the demo app:
```
cellAlign='center'
slidesToShow={1.25}
wrapAround
```

### LINKED ISSUE #340: NEXT button / infinite loop
is there a way to make the NEXT button nonfunctional after the last slide is displayed, because right now my carousel has a bunch of blank slides after the last slide with material. Or have we figured out the infinite loop yet? wrapAround doesn't work like an infinite loop , or it could be those blank slides ?

mariano-formidable:
I'm re-working some of the wrapAround logic so the current slide always has slides positioned before and after to help prevent slides blinking into existence when cycling through them.

This is a PR will closes #324 , closes #194 , closes #340

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -483,9 +483,9 @@ export default class Carousel extends React.Component {
           prevState => ({
             left: this.props.vertical
               ? 0
-              : this.getTargetLeft(-1, prevState.currentSlide),
+              : this.getTargetLeft(0, prevState.currentSlide),
             top: this.props.vertical
-              ? this.getTargetLeft(-1, prevState.currentSlide)
+              ? this.getTargetLeft(0, prevState.currentSlide)
               : 0,
             currentSlide: endSlide,
             isWrappingAround: true,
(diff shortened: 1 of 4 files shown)


src/index.js
```diff
@@ -113,11 +113,40 @@ export default class Carousel extends React.Component {
...
+        return shouldUpdate;
+      })(this.props, nextProps, [
+        'cellSpacing',
+        'vertical',
+        'slideWidth',
+        'slideHeight',
+        'heightMode',
+        'slidesToScroll',
+        'slidesToShow'
+      ]);
```
samwhale: This is super sweet. I like how it checks if it actually needs to update. easy to manage later too!

src/index.js
```diff
@@ -850,42 +879,54 @@ export default class Carousel extends React.Component {
...
 
-      if (this.state.slideWidth !== 1) {
-        slidesAfter = Math.ceil(
-          (Math.abs(positionValue) - this.state.slideWidth) /
-            this.state.slideWidth
-        );
+      if (direction < 0) {
+        const temp = slidesBefore;
+        slidesBefore = slidesAfter;
+        slidesAfter = temp;
```
samwhale: This totally fixes #194 and #340. Good job getting the bigger half of the slides to show in the direction that it's sliding.

webfrontzhifei:
It will blink when the last slide cycles to the first. so does it not work?

PR #392, opened 2018-08-06
use nextProps in goToSlide to properly scroll to dynamically added slide

sarmeyer:


duongthienlee:
If I slideIndex is not equal to 0, the carousel still can not scroll to next slide. 
However, if I set slideIndex equal to 0 then the problem fixed.
But I want to set inital slideIndex to 3, and then when dynamically adding new slides, the carousel should scroll to next slide

PR #398, opened 2018-08-10
minHeight prevents slides from recalculating their height

### LINKED ISSUE #353: minHeight: 100% leads to incorrect height of the carousel on resize
Slides have a minHeight set to 100% (https://github.com/FormidableLabs/nuka-carousel/pull/334/files).
Due to this change, carousel or slides don't recalculate their height correctly on resize.

**How to reproduce the issue:**
* Start the demo app
* Open http://localhost:8080/ on a wider screen
image
* Resize the window (decrease it's width)
  * **Issue:** The height of the carousel remains the same:
image

**Note:** when increasing the width of the window, it works as expected - carousel height is set properly.

### LINKED ISSUE #394: Incorrect calculating height with vertical view
When it is initialized he shows correctly, but if we try to resize browser window calculation will be wrong

**How to reproduce:**
Try to resize browser window at here https://codesandbox.io/embed/521q9qnmwn

sarmeyer:
On window resize, the minHeight: 100% property was preventing the correct calculation of slide height. I tested this with `heightMode: max` and `vertical: true` props. Slides are responsive to window resizing and slides with different heights are not affected.

fixes #353, fixes #394

### DIFF
--- a/src/transitions/fade-transition.js
+++ b/src/transitions/fade-transition.js
@@ -72,7 +72,6 @@ export default class FadeTransition extends React.Component {
       verticalAlign: 'top',
       width: this.props.slideWidth,
       height: 'auto',
-      minHeight: '100%',
       boxSizing: 'border-box',
       MozBoxSizing: 'border-box',
       marginLeft: this.props.cellSpacing / 2,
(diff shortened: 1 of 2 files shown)


ryan-roemer:
@carlospaelinck -- If you get a chance next week can you verify? Thanks!!!

Issue #313, opened 2018-03-22
V4.0 Could u add new props like `withoutControls`


rkovalyov:
For avoiding awful declaration
```js
<Carousel
  renderTopCenterControls={()=>null}
  renderCenterLeftControls={()=>null}
  renderCenterRightControls={()=>null}
  renderBottomCenterControls={()=>null}
/>
```

Klaasvaak:
Maybe there should be 1 render prop for controls. And let the implementer figure out where to position the controls. 

An other options could be not having control render props at all and make sure the implementer can make these by theirselves using the other callbacks. So this library can focus on only doing the carousel part correctly, which it also does btw 👍

davidbucka:
Hey guys, I just checked in a new pull request that adds the desired prop `withoutControls` 🍻

PR #404, opened 2018-09-01
Add new prop withoutControls

davidbucka:
Hide all controls at once with the prop `withoutControls` as mentioned in #313 .


ashika01:
Thank you for contributing @davidbucka. I have created a new PR with some refactoring and tests.
It can be tracked at PR #409.

sarmeyer:
fixed in PR #409

PR #415, opened 2018-09-18
Fixes #410 wrap around glitch

### LINKED ISSUE #410: Infinite slider wrapAround is not working
By default, the infinite scrolling breaks after reaching the end of the first time. 

I'm using the slider in a Gatsby instance. Happy to provide more details if I can help.

ashika01:
Fixes #410 

- Refactoring setDimensions which separates SlideHeight and SlideWidth logics into a function.
- Set calculated height and width in separate function.

Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -67,6 +67,7 @@ export default class Carousel extends React.Component {
       frameWidth: 0,
       left: 0,
       slideCount: 0,
+      slideHeight: 0,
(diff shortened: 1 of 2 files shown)


ashika01:
Added end to end check for testing wrapAround and setLeft logics.

cc: @sarmeyer

Issue #421, opened 2018-10-02
nuka-carousel stops click events from hitting child elements if 'dragging' prop is set to false


JWroe:
### Describe Your Environment
- nuka-carousel version? **4.4.0**
- React? **16.5.0**
- Browser? **Chrome  69.0.3497.100**

### Describe the Problem
As per the title. If dragging is true I can click my button. If dragging is false I can't click the button. Snippet is below. 
```
        <Carousel 
          dragging={false}
          withoutControls={true}
        >
          <button onClick={() => alert('I was clicked')} />
        </Carousel>
```

How do I work around this so that users can't drag but can click? 

Thanks,

Jamie

jflayhart:
Yup... https://github.com/FormidableLabs/nuka-carousel/issues/399

jflayhart:
@sarmeyer https://github.com/FormidableLabs/nuka-carousel/pull/423

sarmeyer:
fixed in #423

Issue #399, opened 2018-08-14
onClickCapture event handler prevents slide onClick events from triggering (when dragging is disabled)


Francesco-Lanciana:
Bug report

**What is the current behavior?**
When children of carousel have an onClick event handler it won't be triggered unless you have set the dragging option to true.

**If the current behavior is a bug, please provide the steps to reproduce.**

Clicking the div in the Carousel rendered by the following React component will not trigger the console.log:
```
import React from "react";
import Carousel from 'nuka-carousel';

const BrokenCarousel = () => (
    <Carousel dragging={false}>
        <div onClick={() => { console.log("What does it matter, I won't appear anyway") }} style={{width: "100px", height: "100px", background: "blue"}}/>  
    </Carousel>
)

export default BrokenCarousel;
```

If you trigger a touch event (by changing the screen to a phone) then you will see the log appear in the console. The only way to circumvent this (that I have discovered) is to set draggable to true.

**What is the expected behavior?**
The onClick handler should be invoked regardless of what draggable is set to.

Either the onClickCapture should be removed to allow the onClick handler of any children of the carousel to be invoked before the onClick handler of the carousel itself.
OR the event propogation should not be stopped within the handleClick method
OR something else that prevents this problem.

**Other relevant information:**
The reason it works when dragging is set to true is because of the following:
* onMouseUp is invoked before onClick.
* onMouseUp invokes handleSwipe which sets this.clickSafe to false in the event that there was no movement (which there wasn't because we only clicked). 
* handleClick now gets invoked but this.clickSafe is set to false so propagation of the current event is not halted. 

The following commit triggered this behaviour:

[fix extra click event triggered when dragging slide](https://github.com/FormidableLabs/nuka-carousel/commit/860f2db54f82aa010edefa4769748d485eeab7ef#diff-1fdf421c05c1140f6d71444ea2b27638)

minipai:
👍

yantonsoup:
Bump on this, causing lots of issues over here.

benjamminj:
Just came up on this issue today, would love to do a fix if anyone wants to point me in the right direction.

Until we get this fixed, [here's an CodeSandbox](https://codesandbox.io/s/p49nmjwl7) of my temporary workaround. The gist is basically is leave the carousel with `draggable={true}` mode and intercept / cancel the mouse-move events on my own.

lukeliasi:
[...]

PR #423, opened 2018-10-07
Prevent click events only when swiping

jflayhart:
### Description

Great library, but it is taking too much responsibility for handling click events than it should and is therefore causing side-effects for users of this library. For example,

https://github.com/FormidableLabs/nuka-carousel/issues/421
https://github.com/FormidableLabs/nuka-carousel/issues/399
https://github.com/FormidableLabs/nuka-carousel/pull/346
https://github.com/FormidableLabs/nuka-carousel/issues/168
https://github.com/FormidableLabs/nuka-carousel/pull/190
https://github.com/FormidableLabs/nuka-carousel/pull/266
https://github.com/FormidableLabs/nuka-carousel/pull/200

Fixes issue: https://github.com/FormidableLabs/nuka-carousel/issues/168

First off, I don't think a library like this should EVER swallow all click events from happening! I mean, this literally kills all click events unless we want dragging to be true. This is not a good DX nor UX for endusers of any kind. It is also causes indeterministic behavior (see above issues).

Therefore, why not just disable/prevent click events only when swiping?

#### Type of Change

Simple class property `clickDisabled` that disables click events when swiping.

- [x] Bug fix - Improved experience for developers and end users by only disabling click events when swiping (AKA if `dragging` is true).
- [x] Breaking Change - Because this re-enables propagation and does NOT prevent default browser click behavior, this could affect users of this library who are not properly handling their own event propagation.

This should NOT cause any regressions as far as https://github.com/FormidableLabs/nuka-carousel/pull/220 is concerned. Also, only the developers of an application that use this library know how to handle their own click events; this library should not try to account for click behavior by disabling click events altogether.

### How Has This Been Tested?

`yarn run check && yarn run test-e2e`

I also tested this within our own web app and the demo of this library. The idea is that we disable all click events ONLY when dragging/swiping, then reenable `clickSafe` behavior after the swiping is done.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -37,7 +37,7 @@ export default class Carousel extends React.Component {
     super(...arguments);
 
     this.displayName = 'Carousel';
-    this.clickSafe = true;
+    this.clickDisabled = false;
     this.touchObject = {};
     this.controlsMap = [
       { funcName: 'renderTopLeftControls', key: 'TopLeft' },
@@ -345,7 +345,7 @@ export default class Carousel extends React.Component {
               Math.sqrt(Math.pow(e.clientX - this.touchObject.startX, 2))
             );
 
-        this.clickSafe = false;
+        this.clickDisabled = true;
         this.touchObject = {
           startX: this.touchObject.startX,
           startY: this.touchObject.startY,
@@ -412,7 +412,7 @@ export default class Carousel extends React.Component {
   }
 
   handleClick(event) {
-    if (this.clickSafe === false) {
+    if (this.clickDisabled === true) {
       event.preventDefault();
       event.stopPropagation();
 
@@ -448,9 +448,9 @@ export default class Carousel extends React.Component {
       this.goToSlide(this.state.currentSlide);
     }
 
-    // wait for `handleClick` event before resetting clickSafe
+    // wait for `handleClick` event before resetting clickDisabled
     setTimeout(() => {
-      this.clickSafe = true;
+      this.clickDisabled = false;
     }, 200);
     this.touchObject = {};
 
--- a/src/index.js
+++ b/src/index.js
@@ -345,6 +345,7 @@ export default class Carousel extends React.Component {
               Math.sqrt(Math.pow(e.clientX - this.touchObject.startX, 2))
             );
 
+        this.clickSafe = false;
         this.touchObject = {
           startX: this.touchObject.startX,
           startY: this.touchObject.startY,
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -345,6 +345,7 @@ export default class Carousel extends React.Component {
               Math.sqrt(Math.pow(e.clientX - this.touchObject.startX, 2))
             );
 
+        this.clickSafe = false;
```
jflayhart: it is NOT "click safe" when we are dragging slides with mouse

src/index.js
```diff
@@ -411,7 +412,7 @@ export default class Carousel extends React.Component {
   }
 
   handleClick(event) {
-    if (this.clickSafe === true) {
+    if (this.clickSafe === false) {
```
jflayhart: since it is not `clickSafe` or safe to click, disable all click events.

src/index.js
```diff
@@ -421,15 +422,6 @@ export default class Carousel extends React.Component {
...
   handleSwipe() {
-    if (
-      typeof this.touchObject.length !== 'undefined' &&
-      this.touchObject.length > 44
-    ) {
-      this.clickSafe = true;
-    } else {
-      this.clickSafe = false;
-    }
-
```
jflayhart: These lines seemed irrelevant now.

sarmeyer:
Hey @jflayhart thanks for this PR. I agree -  Nuka is taking too much responsibility for handling click events and as a reusable library, we should leave some responsibility up to the users in order to not cause unwanted side effects for other users. I appreciate you taking the time to submit this PR and of course for using Nuka!

sarmeyer:
FYI @jflayhart pushed release 4.4.1 with this fix in it

jflayhart:
Thanks alot @sarmeyer!

onin-hub:
//this work on me. have a try.

//Set usestate:

  const [isCarouselDragging, setIsCarouselDragging] = useState<boolean>(false);
 
 //Set Function:
 
   const handleCarouselDragEnd = () => {
    setTimeout(() => {
      setIsCarouselDragging(false);
    }, 300);
  };

  const handleOnDrag = () => {
    setIsCarouselDragging(true);
  };
  
 // Nuka Component:
  
  <NukaCarousel
     onDrag={handleOnDrag}
     onDragEnd={handleCarouselDragEnd}
  >
  <div onClick={() => {
                              if (!isCarouselDragging) {
                                mapDataCollection(mapItem);
                              }
                            }}>
                            </div>
  </NukaCarousel>

PR #425, opened 2018-10-10
 Feature/accessibility

### LINKED ISSUE #197: Accessibility support
Does we have the Accessibility support built in? Am seeing the implementation is not passing the accessibility requirement. Like the tab order ha to be previous button then the visible slides then the next button and so on.

ElreyB:
### Description

Adding accessibility to carousel
- [x] Adding keyboard controls
- [x] Reader lets you know what slide you are on
- [x] Slide that is being display will be read by reader 

Fixes #197

#### Type of Change

### How Has This Been Tested?

Manually test using ChromeVox and WAVE Evaluation Tool


ElreyB:
@sarmeyer I was able to add some accessibility to nuka-carousel except a `pause` keyboard control. There are a couple of things control the autoplay which didn't alone for `pause` control by the keyboard. But from what I can tell everything should be good. Let me know what you think.

src/index.js
```diff
@@ -1150,7 +1150,16 @@ export default class Carousel extends React.Component {
                 deltaX={tx}
                 deltaY={ty}
               >
-                {validChildren}
+                {React.Children.map(validChildren, (child, index) => {
+                  const ariaProps =
```
ElreyB: @sarmeyer So I am using the same logic as before here but instead of using css `{display: none}` I am using `{aria-hidden: true}`. This way the images are still seen on the DOM keeping  the look it had before but won't be read my the reader. Also, the tabbing order got messed up because I added `tabindex: 1` to all the images. Now its just adding `tabindex` the one the is visible.

src/index.js
```diff
@@ -477,6 +478,16 @@ export default class Carousel extends React.Component {
       case 40:
         this.previousSlide();
         break;
+      case 32:
+        if (this.state.pauseOnHover && this.props.autoplay) {
```
ElreyB: @sarmeyer I tested this out but since you know a little more of the use cases for the carousel let me know if it actually works.

src/index.js
jflayhart: @ElreyB So I just updated to this version and narrowed down a bug to this change that is highlighted here.

On mobile iPhone devices VoiceOver correctly focuses and reads the FIRST slide, but then skips all the rest of the visible slides. Instead it goes to the next div (i.e. some other button on the page) effectively causing a11y users to miss out on all the other slides.

Any ideas on what needs to be fixed here?

Issue #427, opened 2018-10-15
when click next button quickly, there is infinite animation


DualWield:
# Bugs and Questions

### Prerequisites

**Feel free to delete this section if you have checked off all of the following.**

- [x] I've searched open [issues](https://github.com/FormidableLabs/nuka-carousel/issues) to make sure I'm not opening a duplicate issue
- [x] I have read through the docs before asking a question
- [x] I am using the latest version of nuka-carousel

### Describe Your Environment

- What version of nuka-carousel are you using?
4.4.1
- What version of React are you using?
16.5.2
- What browser are you using?
chrome
- What machine are you on?
mac

### Describe the Problem
When click next button quickly twice, the slider move around, never stop.
https://codesandbox.io/s/2vnmkr2wry

**Expected behavior:**  quickly move to index 2 slide.
**Actual behavior:**  move around and never stop
![test](https://user-images.githubusercontent.com/4132470/46947148-8048b100-d0ac-11e8-9042-96cfec068ff9.gif)

dennishansen:
Also experiencing this issue. Sort of a deal breaker :/

ElreyB:
@DualWield thanks for bringing this up. What is happening here is that react is batching the state of your App component and the state of nuka-carousel. React does this for performance optimization. So when you tap the next button quickly twice your App component and nuka-carousel component slideIndexes are not sync. For this case you would want to update asynchronously as mention in the React docs

https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous

```
<Carousel
        slideIndex={this.state.slideIndex}
        afterSlide={slideIndex => this.setState((prevState) => ({slideIndex: prevState.slideIndex + 1}))}
      >
```

However, since nuka-carousel applies `this.props.slideIndex` to `this.state.currentSlide` internally I would recommend this solution

```
  <Carousel
        slideIndex={this.state.slideIndex}
        afterSlide={slideIndex => this.setState({ currentSlide: slideIndex })}
      >
```
I have notice that in the example in the README that it is not written this way. We will look into update the README to reflect this or code base to allow `this.setState({slideIndex})`.
Let us know if this was helpful. Happy coding!

CharlyJazz:
I think this needs a solution inside the internal nuka API. It's a weird behavior that nobody wants to have on their web pages.

PR #429, opened 2018-10-18
update external control of carousel state example

ElreyB:
### Description

Notice there were a lot of issues surrounding external controls and the `afterSlide` props function. 
```
  <Carousel
        slideIndex={this.state.slideIndex}
        afterSlide={slideIndex => this.setState({ slideIndex })}
      >
  ```

But in the code base there is nothing looking for an external `this.state.slideIndex`. So now React is batching two states that are out of sync with each other. I believe the README should be like this.

```
  <Carousel
        slideIndex={this.state.slideIndex}
        afterSlide={slideIndex => this.setState({ currentSlide: slideIndex })}
      >
```
I this way the external slideIndex is in sync with nuka-carousel's state. This should fix all issues surrounding `afterSlide`. If nuka-carousel  is suppose to recognize the external slideIndex some refactoring of the code base will be needed. 

### How Has This Been Tested?

Manually tested 

- [ ] My code follows the style guidelines of this project (I have run `yarn lint`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ export default class extends React.Component {
     return (
       <Carousel
         slideIndex={this.state.slideIndex}
-        afterSlide={slideIndex => this.setState({ slideIndex })}
+        afterSlide={slideIndex => this.setState({ currentSlide: slideIndex })}
       >
         <img src="http://placehold.it/1000x400/ffffff/c0392b/&text=slide1" />
         <img src="http://placehold.it/1000x400/ffffff/c0392b/&text=slide2" />


sarmeyer:
@ElreyB Does this address any specific tickets? Would you mind linking them here?

ElreyB:
@sarmeyer For #427 but to be honest I think this change is not needed. If you look at the issue comments people expect it to be part of the nuka-carousel api which makes sense.

sarmeyer:
I'll close this since it seems you have a good working solution in progress

PR #435, opened 2018-10-25
Fix issues where `clickDisabled` is set to true too often

jflayhart:
### Description

Endusers of my implementation of nuka have been complaining that sometimes their clicks don't register. So I looked into this...

1) I found that 200ms is a bit too high to wait before resetting `clickDisabled` state after swiping. This caused some users' clicks not to register even after the swiping behavior. Clicks should only be disabled for the minimum amount of time necessary while swiping/dragging and not a moment later.
2) Occasionally `onMouseMove` is invoked even at the slightest mouse movement. This is unnoticeable to the user, but noticeable to the code which then disables the click event altogether leaving the enduser confused.

#### Type of Change

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -458,7 +458,7 @@ export default class Carousel extends React.Component {
     // wait for `handleClick` event before resetting clickDisabled
     setTimeout(() => {
       this.clickDisabled = false;
-    }, 200);
+    }, 0);
     this.touchObject = {};
 
     this.setState({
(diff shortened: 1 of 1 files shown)


jflayhart:
@sarmeyer I may have introduced a minor bug with [my last PR](https://github.com/FormidableLabs/nuka-carousel/pull/423). Please review. Thanks!

PR #436, opened 2018-10-25
Chore/refactor to diff files

ElreyB:
### Description

Separated functions that could be use as pure function instead of class functions. Reducing `index.js` file from 1290 lines to 1023 lines. Since all the functions were tested through the class I didn't make an extra test file for the additional files.

#### Type of Change

- [x] Move functions to separate files

### How Has This Been Tested?

yarn test && yarn test-e2e


src/index.js
```diff
@@ -599,9 +573,19 @@ export default class Carousel extends React.Component {
...
-              : this.getTargetLeft(0, prevState.currentSlide),
+              : getTargetLeft(
+                  0,
+                  prevState.currentSlide,
+                  this.props,
+                  this.state
+                ),
             top: props.vertical
-              ? this.getTargetLeft(0, prevState.currentSlide)
+              ? getTargetLeft(
```
ashika01: I would set this to '0' considering we know the value type

src/utilities/style-utilities.js
```diff
@@ -0,0 +1,176 @@
...
+};
+
+export const getOffsetDeltas = (touchObject, props, state) => {
+  const { vertical } = props;
+  const { isWrappingAround, wrapToIndex } = state;
+  const { length, direction } = touchObject;
+  let offset = 0;
+
+  if (isWrappingAround) {
+    offset = getTargetLeft(null, wrapToIndex, props, state);
```
ashika01: Again here, knowing the value type, better to use '0'

src/utilities/animation-utility.js
```diff
@@ -0,0 +1,41 @@
+export const getTargetLeft = (touchOffset, slide, props, state) => {
```
ElreyB: @ashika01 @sarmeyer Me and Ashika were talking about moving this function back to Carousel class since it is used so many times in the class component. After looking it over a bit and asking around I would like to keep it in this `animation-utility.js` file. This function is taken in some arguments and returning a calculation based off those arguments. It doesn't need to belong to the class itself. In theory, this function could calculate the same thing for any class component that would need it with out know about the class component itself.  Since the main purpose for this branch is to move as many function as possible to a separate file to make the `index.js` readable I would like to keep this in the `animation-utility.js` file. I would like to another option about this and would like to hear the best practice for something like this.

src/utilities/utilities.js
```diff
@@ -54,3 +54,60 @@ export const addAccessibility = (children, slidesToShow, currentSlide) => {
     });
   }
 };
+
+export const getPropsByTransitionMode = (props, keys) => {
+  const { slidesToShow, transitionMode } = props;
```
sarmeyer: you can destructure these props in the function declaration like so: ```export const getPropsByTransitionMode = ({slidesToShow, transitionMode}, keys) => {...```

sarmeyer: if we only need those properties, it's clearer for the reader to see which ones we need explicitly, rather than getting all of props

ElreyB: Awesome. Will do.

ElreyB: Moving the destructuring to the function declaration won't work for the `default` and `else` statement of the function. I tried a couple of different things but it always break the function.

sarmeyer: Ah I see - yes, we'll need the whole props object to access the key properties on it. so we won't be able to destructure the properties we need from props in the first line. good catch!

PR #445, opened 2018-11-08
add keyboardControls prop to disable keyboard controls

### LINKED ISSUE #443: Disable keyboard shortcuts
# Bugs and Questions

Is it possible to disable keyboard shortcuts? If not, I'd be happy to create a PR to add something like a `disableShortcuts` prop.

In our case, the carousel is part of a page that contains forms. When typing in the form, the carousel starts going crazy.

### Describe Your Environment

- What version of nuka-carousel are you using?
4.4.2
- What version of React are you using?
16.6.0
- What browser are you using?
Chrome 70
- What machine are you on?
Mac

ElreyB:
### Description
Add `keyboardControls` prop to disable `nuka-carousel` keyboard controls

Fixes #443

#### Type of Change

### How Has This Been Tested?

yarn test && yarn test-e2e


README.md
```diff
@@ -79,6 +79,7 @@ Or on CodeSandeBox at the following url: <a href="https://codesandbox.io/s/mjxwl
...
 | initialSlideHeight   | `React.PropTypes.number`                                                                | Initial height of the slides in pixels.                                                                                                                                                       | `100`               |
 | initialSlideWidth    | `React.PropTypes.number`                                                                | Initial width of the slides in pixels                                                                                                                                                         |                     |
+| keyboardControls     | `React.PropTypes.bool`                                                                  | When set to `false` will disable keyboard controls.                                                                                                                                           | `true`              |
```
sarmeyer: I would rename this to `disableKeyboardControls` instead. I think it makes it more clear what this prop is doing

sarmeyer: You'll have to switch the bool values in this case

test/specs/carousel.test.js
```diff
@@ -533,6 +533,42 @@ describe('<Carousel />', () => {
...
+            <p>Slide2</p>
+            <p>Slide3</p>
+          </Carousel>
+        );
+        expect(wrapper).toHaveState({ currentSlide: 0 });
+        map.keydown({ keyCode: 39 });
+        expect(wrapper).toHaveState({ currentSlide: 1 });
+      });
+
+      it('should not move to next slide when pressing right arrow key', () => {
```
sarmeyer: Include in the test description what prop is changing so we know why it shouldn't be moving to the next slide in this case

test/specs/carousel.test.js
```diff
@@ -601,6 +637,42 @@ describe('<Carousel />', () => {
 
         expect(defaultWrapper).toHaveState({ cellAlign: 'left' });
       });
+
+      it('should move to next slide when pressing right arrow key', () => {
```
sarmeyer: it looks like these tests are in here twice?

test/specs/carousel.test.js
```diff
@@ -474,6 +474,44 @@ describe('<Carousel />', () => {
...
+            <p>Slide2</p>
+            <p>Slide3</p>
+          </Carousel>
+        );
+        expect(wrapper).toHaveState({ currentSlide: 0 });
+        map.keydown({ keyCode: 39 });
+        expect(wrapper).toHaveState({ currentSlide: 1 });
+      });
+
+      it('should not move to next slide when pressing right arrow key', () => {
```
sarmeyer: this is just a nit picky thing - but it might be nice to include the name of prop that is changing the behavior in the assertion so that the reader knows why it should not move to the next slide in this case. something like `should not move to next slide when pressing right arrow key when disableKeyboardControls is set to true`

j0nd0n7:
When using in forms, It would be nice to have the possibility to use the keyboard only when the carousel "has the focus". If I'm not wrong https://github.com/akiran/react-slick has something like this.

j0nd0n7:
Also, nice PR work!! Waiting for this to be merged!

test/specs/carousel.test.js
```diff
@@ -494,22 +494,24 @@ describe('<Carousel />', () => {
         expect(wrapper).toHaveState({ currentSlide: 1 });
       });
 
-      it('should not move to next slide when pressing right arrow key', () => {
-        const map = {};
-        document.addEventListener = jest.fn((event, cb) => {
-          map[event] = cb;
+      describe('when disableKeyboardControls is set to true', () => {
```
sarmeyer: sorry to make this confusing! we don't need another `describe` block, I just meant the assertion (`it...does some thing`) was more descriptive.

sarmeyer: so it would look like `it('should not move to next slide when pressing right arrow key when disableKeyboardControls is set to true', () => {`

sarmeyer:
@j0nd0n7 that's a great idea. I'd like to get this PR in first and do that separately. Would you like to submit a PR once this is merged/released today?

src/index.js
```diff
@@ -400,35 +400,37 @@ export default class Carousel extends React.Component {
   }
   // eslint-disable-next-line complexity
   handleKeyPress(e) {
```
sarmeyer: instead of checking for the `disableKeyboardControls` prop inside of this function, you can check for it before the method is even called. That way, the method isn't called if `disableKeyboardControls` is false

sarmeyer: on lines 695 and 720

ElreyB: @sarmeyer I would be worry that this could cause a potential bug.  If the user changes `disableKeyboardControls` at any point it may not get caught right away, depending on how the user changes the prop. If we move the check on the mount we will also need to add / remove / update the event listener during the `componentWillReceiveProps`. Alternatively, we won't be listening to changes to that prop. Since we are trying to simplify `componentWillReceiveProps` as part of the refactor I would consider leave the check in the `handleKeypress` event.

khawerrind:
Any plan on merging this PR?

test/specs/carousel.test.js
sarmeyer: has this been resolved? @ElreyB

src/index.js
sarmeyer: Yea, I understand. We can leave the check where it is for now

PR #465, opened 2018-12-11
Add touch-action style only when swiping is true

sjfkai:
### Description

We don't need `touch-action` on frame when `swiping` is false.
That will prevent scroll when touch upon `Carousel` element.

You can test at [https://codesandbox.io/s/o93438q32y](https://codesandbox.io/s/o93438q32y) with mobile device mode.

- [ ] I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -934,7 +934,8 @@ export default class Carousel extends React.Component {
       framePadding,
       slidesToShow,
       renderAnnounceSlideMessage,
-      disableAnimation
+      disableAnimation,
+      swiping
     } = this.props;
     const duration =
       this.state.dragging ||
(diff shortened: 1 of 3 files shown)


sarmeyer:
@sjfkai How do you intend `swiping` to be set? What is your specific use case for this? 

I can see how it would be easy to get "stuck" in a vertically scrolling carousel on a mobile device, if the width of the carousel is 100% of the viewport. I'm debating whether this is something the carousel component should handle.

sjfkai:
@sarmeyer I'm sorry for the wrong title.  What I mean is we don't need `touch-action` when swiping is false.

My use case is I don't need change slide by `swiping`. I only want it auto play all the time. In this case, if I `swiping` on the  carousel component, I want the parent element to scroll.

But the `touch-action` style blocked the parent element scrolling.

PR #466, opened 2018-12-11
Prevent multiple afterSlide calls of which some used undefined variabes, breaking slide-changes

pimverlangen:
### Description

After upgrading Nuka Carousel to 4.4.4, our `afterSlide` callbacks were being called twice and with `slideIndex === NaN`. After quickly checking, it seemed that the library tries to call `props.slideCount`, but `slideCount` is not a prop of the component. Further investigating, seemed that the before/after callbacks seemed to be best placed in the `goToSlide` function, so that's what this PR also includes (`afterSlide` is not used outside of this function anymore).

#### Type of Change

### How Has This Been Tested?

Switching slides and adding a console.log to the `afterSlide` prop to see how often it gets called and with what value. Then debugged it down some more to come to the fixes in this PR.

### Checklist: 

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -561,6 +561,7 @@ export default class Carousel extends React.Component {
 
     this.setState({ easing: easing[props.easing] });
     this.isTransitioning = true;
+    const previousSlide = this.state.currentSlide;
 
     if (index >= this.state.slideCount || index < 0) {
       if (!props.wrapAround) {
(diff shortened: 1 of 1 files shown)


wenJonSnow:
How to solve this problem?

wenJonSnow:
I have encountered this problem. How can I solve it?

sjfkai:
> I have encountered this problem. How can I solve it?

@wenJonSnow  You can try to use the previous version, like v4.4.1.

wenJonSnow:
@sjfkai  Thank you, but version 4.4.1 will shake the page when it slides.

fi3ework:
please merge this

sarmeyer:
Will have this in a release by the end of the week

Issue #457, opened 2018-11-21
Initial Height too long (broken)


rockmandash:
Initial Height too long (broken)
`initialSlideHeight` & `heightMode` both not helping.
Have to wait until next slide, the height will back to normal.

willmoss1000:
I am getting this too.

rayyapari:
I have same issue, need help :(

haugan:
> 
> 
> Duplicate of: #202 @sarmeyer / @rockmandash please close.

The duplicate is closed, so closing this new one won't help anyone. The problem needs to be addressed.

vyteniskuciauskas-home24:
Reproducible bug https://codesandbox.io/s/q83z6q0636

sarmeyer:
fixed with #469

devedupereira:
Please, reopen this issue since initial height isn't fixed well.

sarmeyer:
@vyteniskuciauskas-home24 I'm not able to reproduce this issue. Is anyone able to send me an example with this issue present?

vyteniskuciauskas-home24:
works for me also

devedupereira:
@sarmeyer and @vyteniskuciauskas-home24 For me, the first render is with long height and after resize it normalize with correct height value.

sarmeyer:
@vyteniskuciauskas-home24 this is what I see when the codesandbox loads: 
![image](https://user-images.githubusercontent.com/18433259/50903415-5a008c00-13e3-11e9-8cca-28b516b5fe99.png)

The height looks correct, and it doesn't change as I move through the slides. Is there another example you have?

devedupereira:
@sarmeyer look these screenshots:

before resize, slider apply 5982px of height
https://i.imgur.com/sxV9DXx.png

after resize, everything work fine:
https://i.imgur.com/KMNRdZw.png

sarmeyer:
@devedupereira what version of nuka are you using? Thanks for helping me pin this down, I'm still not able to reproduce it locally, so your screenshots help!

devedupereira:
@sarmeyer latest version 4.4.5

Caldwerl:
I will add that this is occurring for myself in V 4.4.5, not on first page load, but after I navigate into a view with a carousel.
First page load is fine, perfect, if I navigate to a page with a carousel, it has over double the height until the next slide.

So most likely you wont be recreating this on a single view version.

[...]

Issue #417, opened 2018-09-22
Initial height is broken


balazsorban44:
### Describe Your Environment

- What version of nuka-carousel are you using? latest
- What version of React are you using? latest
- What browser are you using? chrome
- What machine are you on? linux

### Describe the Problem

**Expected behavior:** [What you expect to happen]
That the initial height is the first (or the tallest) image's height
**Actual behavior:** [What actually happens]
No matter what I set, the initial height is 100px
Example:
[![Edit o58601znvq](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/o58601znvq)
### Additional Information
I wrote about this at the issue #413 but I move it into a new issue for clearance.

> When the carousel first loads, it is only 100px high, no matter what I set where. Changing to the next slide, or resizing the window, everything snaps into place, and the height becomes the image's height as it is supposed to be.

ashika01:
@balazsorban44 : we absorb the behavior only in Sandbox and not when you install and run the package in a browser. Could you let us know if you face the issue outside CodeSandbox?

dennis3124:
Same issue, the initial height is always miscalculated for some reason. Once i slide to the next slide, it readjusts correctly

synekdok:
Might be useful for some - here's a veeeery hacky workaround I've been using to get the carousel to the correct height without having to interact with it first

```
componentDidMount(){
    setTimeout(() => {
        try {
            this._carousel.setDimensions(null)
        }
        catch(err){
                // sometimes it crashes in chrome, pristine error handling here
        }
    }, 200);
}
```

balazsorban44:
@synekdok Since more of us has the same problem, I don't feel that we should make hacks to solve it, rather implement the solution directly in this package!

NubSoup:
Also, when using in modal position fixed, height/width 100%.  and using image ratio % trick.  slides are rendering at 20px height on first load until resize is triggered. Also it only resizes properly when it's the width that changes in the resize, not height.  

need to address resizing height properly when height changes and width does not.

example code:
https://codesandbox.io/s/j4n5046k7w

edwinharly:
In my case, the initial height is 0,

[![Edit pkx5p4jj9x](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/pkx5p4jj9x)

ianregister:
[...]

Issue #103, opened 2016-06-14
Slides sometimes render with zero height


aczerepinski:
I'm testing out 2.0.0 via the master branch, and seeing a new issue where slides will sometimes render with zero height. This is unrelated to/not fixed by commit 5ed9ff5. Resizing my browser window adds height to the slides instantly. I added this code to setDimensions, which fixes it for the most part:

```
if (firstSlide.offsetHeight === 0) {
   var width = window.innerWidth;
   firstSlide.style.width = width + 'px';
   firstSlide.style.height = (width/2.4) + 'px'; //known ratio for my images - needed for Safari
}
```

I'm not submitting this as a PR right now because it feels a bit hacky and istied to my external CSS and image ratio. Is anyone else experiencing this? Is there a cleaner solution?

This isn't an issue on ^1.2.0.

kenwheeler:
It's probably the absolute positioning... Necessary for wrapAround, I didn't want to do it, but people need it so I did. What's your fix look like generally? Would love to undo any damage done, sooner than later.

aczerepinski:
Sorry - after further testing, the fix I was working on probably isn't the right direction. It does force the slides to have height, but introduces other drawbacks such as slide transition hiccups.  It I'll try to find something cleaner that I can contribute back, but in the meantime I may need to cherry pick the autoplay code into 1.2 without wrapAround. I might wrap by calling `goToSlide(0)`, and leave the CSS alone.

Also, I'm using plenty of my own CSS, external controllers, etc. If nobody else is chiming in with "me too" perhaps this is just a CSS conflict that won't effect everyone. Feel free to close.

lizlove:
I'm having the same issue with slides rendering with zero height and resizing the browser adds height to the slides instantly as well. I removed all applicable css.  Currently if I set an initialSlideHeight in pixels it will avoid the blank slideshow, but my slides then do not size to the frame (instead overlap, etc). Is there a way to avoid the zero state without compromising the scaling? 
Thanks!

Godefroy:
Same issue here.

1. First load: Height=0, I think it's because the carousel adjust its height when images are not yet loaded.
2. Resize browser: Perfect.
3. Resize again: Width ok, Height not ok (stay the same than after first resize)
4. Second load (remove and add component again without refreshing the page): Perfect.

So there are two problems: size detection on load, and browser resize.

alexanderfrey:
Same problem here for the 2.0.2

[...]

PR #469, opened 2018-12-14
Fix initial height

### LINKED ISSUE #202: setDimensions bug? ul.slider-list too tall after componentDidMount
My horizontal slider with all default props is too tall on first mount, and I think this is why:

* `componentWillMount -> setInitialDimensions` sets `ul.slider-list` and each `li.slider-slide` to zero width

* Then we come to `componentDidMount -> setDimensions`, where

* `firstSlide.style.height = 'auto';` allows the `li.slider-slide` to grow vertically 

* which is then measured to calculate the height of the slider 
`slideHeight = firstSlide.offsetHeight;`

**BUT**

* `li.firstSlide` is still zero width! So all of its content gets squeezed to minimal width, forcing it down and making it way too tall!

* A few lines later, `slideWidth` is properly calculated. 
`slideWidth = frame.offsetWidth / props.slidesToShow * props.slideWidth;`
But it's not applied to the DOM yet, so we still wouldnt be able to properly measure the `firstSlide.offsetHeight`

* Then `setState` triggers a render which applies the correct width to `ul.slider-list` and `li.slider-slide`, but the **wrong height** to `ul.slider-list`. 

* Then, if we're lucky enough to have mounted this component on first page load, we may have a `readystatechange` event trigger another `setDimensions` which will calculate the correct `firstSlide.offsetHeight`, this time because the correct `slideWidth` has already been applied to the DOM.

So basically `setDimensions` needs to run twice *after* mounting in order to get it right. And if I'm mounting this component in a SPA long after readystatechange has fired, it only runs once. Am I doing something wrong? Or is this a bug. I'm not sure what the fix should be because I'm afraid that fixing for my own case would break other configurations. Whatever the case, **the first slide needs to be given its final width in the DOM before its offsetHeight is measured.**

vyteniskuciauskas-home24:
Trying to fix initial height bug
This should fix the following reproducible issue: https://codesandbox.io/s/0qr88707wp

Closes #202 #457 #417 #103

### DIFF
--- a/src/utilities/style-utilities.js
+++ b/src/utilities/style-utilities.js
@@ -98,7 +98,7 @@ export const getSliderStyles = (propWidth, stateSlideWidth) => {
     position: 'relative',
     display: 'block',
     width: propWidth,
-    height: 'auto',
+    height: '100%',
     boxSizing: 'border-box',
     MozBoxSizing: 'border-box',
     visibility: stateSlideWidth ? 'inherit' : 'hidden'
@@ -115,7 +115,7 @@ export const getFrameStyles = (
     position: 'relative',
     display: 'block',
     overflow: propFrameOverFlow,
-    height: propVertical ? stateFrameWidth || 'initial' : 'auto',
+    height: propVertical ? stateFrameWidth || 'initial' : '100%',
     margin: propFramePadding,
     padding: 0,
     transform: 'translate3d(0, 0, 0)',
--- a/src/utilities/style-utilities.js
+++ b/src/utilities/style-utilities.js
@@ -1,6 +1,6 @@
 export const getImgTagStyles = () => {
-  return `.slider-slide > img { width: 100%; display: block;}
-          .slider-slide > img:focus {margin: auto; width: 99.5%; outline-width: 11px}`;
+  return `.slider-slide > img { width: 100%; display: block; }
+          .slider-slide > img:focus { margin: auto; }`;
 };
 
 export const getDecoratorStyles = position => {


sarmeyer:
@vyteniskuciauskas-home24 Just finished testing this and it works great. Thanks for taking the time to submit this PR and doing the work of responding to the issues related. I'll cut a release with this fix in since it's such a needed bug fix!

stefangordon:
Oh wow I have been watching some of those bugs for literally years - almost sad to see them go ;)

Thanks for this!

IMFIL:
Do you know when the release will be live ?

sarmeyer:
@IMFIL release 4.4.5 live today

wereHamster:
I'm still seeing the issue, and it kindof went away when I set `width:100vw` on the slides. I didn't investigate much further.

devedupereira:
I'm still fighting with this error too. @wereHamster your workaround doesn't work fine, sometimes the content inside slide go out from screen because this width setted.

yyynnn:
Same here, height is still 0 when routes change (via browser history).

flikQ:
Doesn't seem to be fixed for me

aamorozov:
Looks like it is still an issue with `v4.4.7`

SummitPatel:
Also still having this issue with `4.4.7`. 

Temporary fix was to wrap my `Review` component (i.e. the slides of the carousel) in a `div` with a height set: 

```
export const ReviewsCarousel = props => (
  <Carousel heightMode={"max"}>
    {props.reviews.length &&
      props.reviews.map((review, i) => {
        return (
          <div style={{ height: 200 }}>
            <Review
              key={i}
              content={review.attributes.review}
              rating={review.attributes.rating}
            />
          </div>
        );
      })}
  </Carousel>
);
```

Blobson:
You can try this [workaround](https://github.com/FormidableLabs/nuka-carousel/issues/457#issuecomment-478349501) to set fixed height and aspect ratio.

jibin2706:
Still having issues with height on initial render

PR #290, opened 2018-03-07
Zoom effect animation

lauterry:
Hi

This pull request allows to have a zoom effect animation while changing slides: 

zoom2

(Sorry for the gif poor quality)

PR #473, opened 2019-01-04
zoom effect

sarmeyer:
Fixed merge conflicts and updated Readme for the following PR
https://github.com/FormidableLabs/nuka-carousel/pull/290


.gitignore
```diff
@@ -34,4 +34,4 @@ dist
 es
 lib
 yarn-error.log*
-package-lock.json
+package-lock.json
```
kale-stew: Does nuka have a lint precommit hook? This eliminated newline seems odd

sarmeyer: There was another line here in the original PR, but it wasn't necessary so I removed it. it was also part of a merge conflict so I'm assuming the reason it isn't showing up in the diff is because of that

src/transitions/scroll-transition.js
```diff
@@ -83,20 +103,25 @@ export default class ScrollTransition extends React.Component {
...
       marginLeft: this.props.vertical ? 'auto' : this.props.cellSpacing / 2,
       marginRight: this.props.vertical ? 'auto' : this.props.cellSpacing / 2,
       marginTop: this.props.vertical ? this.props.cellSpacing / 2 : 'auto',
-      marginBottom: this.props.vertical ? this.props.cellSpacing / 2 : 'auto'
+      MozBoxSizing: 'border-box',
+      position: 'absolute',
+      top: this.props.vertical ? targetPosition : 0,
+      transform:
+        this.props.animation === 'zoom' && this.props.currentSlide !== index
+          ? 'scale(0.85)'
```
coopy: I imagine we'll want to make this scale configurable via a prop in the future. Maybe create an issue and add a  comment here?

src/transitions/scroll-transition.js
```diff
@@ -83,20 +103,25 @@ export default class ScrollTransition extends React.Component {
...
+      marginBottom: this.props.vertical ? this.props.cellSpacing / 2 : 'auto',
       marginLeft: this.props.vertical ? 'auto' : this.props.cellSpacing / 2,
       marginRight: this.props.vertical ? 'auto' : this.props.cellSpacing / 2,
       marginTop: this.props.vertical ? this.props.cellSpacing / 2 : 'auto',
-      marginBottom: this.props.vertical ? this.props.cellSpacing / 2 : 'auto'
+      MozBoxSizing: 'border-box',
+      position: 'absolute',
+      top: this.props.vertical ? targetPosition : 0,
+      transform:
+        this.props.animation === 'zoom' && this.props.currentSlide !== index
```
coopy: This inline logic works fine while there's only one animation type. In the future, I expect we'll want to factor out the transform value resolution to a tested function. For now, this is appropriate! 👍

src/transitions/scroll-transition.js
sarmeyer: good idea

src/transitions/scroll-transition.js
sarmeyer: 👍 sounds good

Issue #474, opened 2019-01-04
Make zoom animation scale configurable via prop


sarmeyer:
Create prop to use in `scroll-transition.js` to configure the `transform` property scale value, when `props.animation === 'zoom'`

PR #488, opened 2019-01-24
Task/configurable zoom animation scale

johannam999:
### Description

Fixes Issue #474

#### Created configurable zoom animation scale and added an input field to Demo to visualize different zoom scale values.

- [x] New feature (a non-breaking change which adds functionality)

### How Has This Been Tested?

Tested zoomScale values in ScrollTransition component.


johannam999:
@sarmeyer @mariano-formidable I think it's ready for a final review. Thnx

test/specs/scroll-transition.test.js
```diff
@@ -53,4 +53,92 @@ describe('<ScrollTransition />', () => {
...
+            <p>Slide 3</p>
+          </ScrollTransition>
+        );
+        const instance = wrapper.instance();
+        expect(instance.getSlideStyles(0, 0).transform).toEqual(`scale(1)`);
+        expect(instance.getSlideStyles(1, 0).transform).toEqual(`scale(0)`);
+        expect(instance.getSlideStyles(2, 0).transform).toEqual(`scale(0)`);
+      });
+
+      it('should return zoomScale of 1 for all slides when animation !== zoom and zoomScale value is below 0 ', () => {
```
sarmeyer: What do you mean by `below 0` here? Are you referring to zoomScale not being passed in? or is the intention to check that the user is passing in a negative value?

johannam999: I was testing that the zoomScale value is not being passed when the animation is not set to "zoom". The positive/negative value does not make any difference in this case. I will change that. Good catch!

mariano-formidable: Yeah, I think you can reword it to " ... and zoomScale is less than 0".

test/specs/scroll-transition.test.js
```diff
@@ -53,4 +53,92 @@ describe('<ScrollTransition />', () => {
...
+            <p>Slide 3</p>
+          </ScrollTransition>
+        );
+        const instance = wrapper.instance();
+        expect(instance.getSlideStyles(0, 0).transform).toEqual(`scale(1)`);
+        expect(instance.getSlideStyles(1, 0).transform).toEqual(`scale(1)`);
+        expect(instance.getSlideStyles(2, 0).transform).toEqual(`scale(1)`);
+      });
+
+      it('should return zoomScale of 1 for slide index 0 and 0.85 as a default for slide index 1 & 2 when animation === zoom but no zoomScale defined ', () => {
```
mariano-formidable: Now that I'm looking at it again.  I think the wording for all these _it_ statements should start with "should return a _scale_ of ...".  

Since `zoomScale` is the name of the prop, it's a little confusing to read:
"should return a zoomScale of 1 if ... zoomScale is [blah]".

But if it said:
"should return a scale of 1 if ... zoomScale is [blah]"

Then I think it's more clear that `zoomScale` is the passed in prop, and `scale` is the calculated result.  Does that make sense?

johannam999: Yes, that makes sense.

test/specs/scroll-transition.test.js
johannam999: Fixed.

test/specs/scroll-transition.test.js
johannam999: Fixed

test/specs/scroll-transition.test.js
```diff
@@ -53,4 +53,92 @@ describe('<ScrollTransition />', () => {
...
+            <p>Slide 3</p>
+          </ScrollTransition>
+        );
+        const instance = wrapper.instance();
+        expect(instance.getSlideStyles(0, 0).transform).toEqual(`scale(1)`);
+        expect(instance.getSlideStyles(1, 0).transform).toEqual(`scale(0.85)`);
+        expect(instance.getSlideStyles(2, 0).transform).toEqual(`scale(0.85)`);
+      });
+
+      it('should return zoomScale of 1 for slide index 0 and zoomScale of 0.1 for index 1 & 2 when animation === zoom and zoomScale = 0.1 ', () => {
```
mariano-formidable: I also think you can be more generic in your wording for your it statements.  

Instead of saying:
"should ... for slide index 0 and ... for index 1 & 2 ..."

You could say:
"should ... for the _current slide_ and ... for the _remaining slides_ ..."

This way if someone is just reading through the tests to see what this prop does, it's clear that it will generically work against any amount of slides and will apply different zoomScales for the `current slide` and the `remaining slides` and not just slide index 0, 1, or 2.

test/specs/scroll-transition.test.js
mariano-formidable: @johannam999 You're right, this test is just to show that zoomScale is ignored, so your it statement can be reworded to something like:
"should ignore zoomScale if animation !== zoom"

And then your test shows how it ignores the 0.5 zoomScale when all the slides are set to a scale of 1 😎

test/specs/scroll-transition.test.js
johannam999: Fixed

test/specs/scroll-transition.test.js
johannam999: Thanks I rephrased that.

sarmeyer:
Tested this out and it all works great - including the last couple changes. You've added good test coverage of the new prop!

PR #492, opened 2019-01-28
`Next` button validation + paging dots sync

johannam999:
### Description

Fixed the issue with `Next` button becoming disabled before the last slide was shown. The button is fully functional with `cellAlign ='left'`. When `cellAlign ='center'` or `'right'`,  `slidesToScroll` prop should be set as default.
Fixed the issue with `Paging Dots` not being synched with the number of pages displayed. Especially when slidesToScroll prop value > 1. This fix covers also the cases when `cellAlign = 'center'` or `'right'`.

Fixes #

#### Type of Change

### How Has This Been Tested?

Thorough testing for both functions.


src/default-controls.js
```diff
@@ -39,15 +39,47 @@ export class NextButton extends React.Component {
...
     super(...arguments);
     this.handleClick = this.handleClick.bind(this);
+    this.nextButtonDisable = this.nextButtonDisable.bind(this);
   }
   handleClick(event) {
     event.preventDefault();
     this.props.nextSlide();
   }
+
+  nextButtonDisable(
```
coopy: This function name (ending with a verb) makes it sound like it's a function with side effects.

I don't have a suggestion for a better name... what do you think?

johannam999: how about `getNextButtonIsDisabled`?

mariano-formidable: I think you're on the right track, but it might be a bit too wordy now.  I think your original method name would work if you changed `Disable` to `Disabled`.   That'll remove the action implication from the name.  

`nextButtonDisabled` reads more like a statement/question (which is good for boolean returning functions), and it even reads well in an if statement (e.g. `if (nextButtonDisabled(...)) { /* do stuff */ }`).

johannam999: Got it, thanks!

src/default-controls.js
```diff
@@ -39,15 +39,64 @@ export class NextButton extends React.Component {
...
     super(...arguments);
     this.handleClick = this.handleClick.bind(this);
+    this.nextButtonDisable = this.nextButtonDisabled.bind(this);
   }
   handleClick(event) {
     event.preventDefault();
     this.props.nextSlide();
   }
+
+  nextButtonDisabled(params) {
```
sarmeyer: I would actually rename this to something like `disableNextButton` - the convention of naming these "action" methods with the verb first helps the reader/future dev understand what the method is doing, while still being able to use the function invocation as a boolean if needed.

src/default-controls.js
```diff
@@ -39,15 +39,64 @@ export class NextButton extends React.Component {
...
+    const {
+      wrapAround,
+      slidesToShow,
+      currentSlide,
+      cellAlign,
+      slideCount
+    } = params;
+
+    let disabledButton = false;
```
sarmeyer: similar to my comment above, I think this variable could be named `disableButton`. It follows the same convention as the function name. As a reader, it's more clear to think if we should "disable button?...yes (true) or no (false)" Does that make sense?

mariano-formidable: I agree with Sarah that if this function were to be named `disableNextButton`, then this variable would read better as `disableButton`.  However, with the function named `nextButtonDisabled`, this variable would read better as `buttonDisabled`.  In both cases it's a shortened version of the method name (simply dropping "next"), and so it reads clearly that this is main boolean of the method.  And also both read better as questions like Sarah mentioned (should we 'disable button?", or is "button disabled?).

src/default-controls.js
mariano-formidable: My only nit with `disableNextButton` (but feel free to ignore!!!) is that it could be read that one could invoke `this.disableNextButton()` and the next button would in fact be disabled (or enabled), but it won't.  This method only returns a boolean which could then be used to disable/enable the next button.  Maybe it should be called `shouldDisableNextButton`, to clearly read that it is only returning a flag to be used and not actually doing the enabling/disabling itself, but that could be overkill in naming :)

test/specs/default-controls.test.js
```diff
@@ -0,0 +1,371 @@
...
+        slidesToShow: 3,
+        currentSlide: 4,
+        cellAlign: 'left',
+        slideCount: 6
+      };
+      expect(instance.nextButtonDisabled(testLeftIndex4)).toEqual(true);
+    });
+
+    // testing extreme scenarios
+    it('should return true for slide index >= 4', () => {
```
sarmeyer: I would add a little more explanation of this test case to the `it` statement, because we don't expect the return value to be true _always_ when `slideIndex >= 4`. So something like, `...when "the other conditions are...etc, etc`. I know this can get a bit wordy, so just see if you could come up with a way to differentiate this test for the edge case it's being used to test

test/specs/default-controls.test.js
```diff
@@ -0,0 +1,371 @@
...
+  describe('#nextButtonDisable for cellAlign = `center`', () => {
+    let instance;
+
+    beforeEach(async () => {
+      const wrapper = mount(<NextButton buttonDisable={() => null} />);
+      instance = wrapper.instance();
+    });
+
+    // testing smaller number of pages
+    it('should return true for slide index 5', () => {
```
sarmeyer: same comment for this `it` statement as the one above. I think it would be more clear why we expect this test case if you add a little more info the statement.

test/specs/default-controls.test.js
```diff
@@ -0,0 +1,371 @@
...
+        slidesToShow: 7,
+        currentSlide: 22,
+        cellAlign: 'center',
+        slideCount: 23
+      };
+      expect(instance.nextButtonDisabled(alignCenterTest4)).toEqual(true);
+    });
+
+    // testing edge case scenarios
+    it('should return true with slide index of 4 when slideCount < slidesToShow', () => {
```
sarmeyer: This is a great example of what I'm trying to convey in the comments above! This test statement is very clear

src/default-controls.js
sarmeyer: Ah true - `disableNextButton` gives the impression that this method is in fact disabling the button, so I will redact my comment...it would make more sense to keep it as it is in that case! this naming thing is hard :)

src/default-controls.js
sarmeyer: I agree - leaving the method name as is, I still think this would be more clear as `buttonDisabled` signifying to the reader that the boolean value of this variable correlates to whether the button is disabled or not. This could just be a variance in how we read it, too

src/default-controls.js
johannam999: Thanks!

sarmeyer:
@johannam999 you're good to merge this in

PR #510, opened 2019-03-06
Feature: autoplay in reverse

david-davidson:
Hi @sarmeyer!!! Hi Formidable!!! 👋🤗👋🤗👋 

### Description
We're happy users of `nuka-carousel` over at @textioHQ. For an upcoming project, we're looking to support slide autorotation _in reverse_, where slide indexes tick down through 5, 4, 3, 2, 1, etc.

This PR proposes a new `autoplayReverse` prop, defaulting to false. When `autoplay` is false, the prop is meaningless. When `autoplay` is true _and_ so is `autoplayReverse`, the new behavior kicks in. 

Not married to this implementation if another approach seems like a better fit!

#### Type of Change

_(☝️Note that I've included the documentation update here; not sure if that unchecks that box?)_

### How Has This Been Tested?

Just playing around with the `yarn start` demo. (It looks like `autoplay` doesn't have programmatic tests, so I didn't attempt them for `autoplayReverse` either.)

/cc @amirahaile @max-winderbaum


ryan-roemer:
Thanks 🕊 !!!

PR #524, opened 2019-04-01
don't wrap handleMouseOut handler in extra function

bpmason1:
### Description

Extra anonymous functions around a component's event handler can prevent event's from being deregistered when the component is unmounted resulting in a memory leak.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -170,12 +170,8 @@ export default class Carousel extends React.Component {
   getTouchEvents() {
     if (this.props.swiping === false) {
       return {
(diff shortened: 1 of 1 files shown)


sarmeyer:
Thanks for submitting this PR - removing these will help with the unnecessary creation of these anonymous functions which is an optimization. I wanted to add that these anonymous functions are created with each render - I don't believe there is a memory leak as they are only in memory for as long as the virtual dom resulting from that render is referenced by react. They are garbage collected once the component is unmounted or a new render occurs.

PR #538, opened 2019-04-30
[DO NOT MERGE] [WIP]upgrading nuka with hooks

ashika01:
### Description

Working on renovating whole repo  and upgrading the carousel with react hooks.

#### Type of Change

#### Comments:
- Still in lot of testing and researching phase
- Working on MVP with most basic operation support and building a skeleton framework


src/new-hooks/components/paging-dots.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import _ from 'lodash';
```
sarmeyer: Just a note as we keep developing this, we don't want to import all of lodash as it is a pretty big package to add to our bundle. So we should only import the methods we are using

sarmeyer: it looks like this `import slice from 'lodash.slice';`

ashika01: Yeah gotcha 👍. I am going for this for now because I am quite unsure on things I would be messing around here. I will update them all pretty soon when I start to clean things up.

sarmeyer: totally! Just wanted to comment on it so I don't forget

PR #547, opened 2019-05-13
Fixed visibility for initial load and added height prop

maurispalletti:
### Description

- Change to show first image in carrousel when loaded, instead of waiting until all assets are loaded. 
- Added prop to set hardcoded height and typescript definition (with default value ```'auto'``` as before).
- Added disableAnimation typescript definition.

#### Type of Change

### How Has This Been Tested?

I have run `yarn test` and `yarn test-e2e` and tests are ok.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -154,6 +154,12 @@ export interface CarouselProps {
   enableKeyboardControls?: boolean;
 
   /**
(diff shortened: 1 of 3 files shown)


sarmeyer:
Looks great - thanks for submitting this PR!

PR #555, opened 2019-06-17
Fix wrongly calculated height for current slide on slide change

karolisgrinkevicius-home24:
### Description

Fix wrongly calculated height for current slide image on slide change
nuka_height_recalc_bug

#### Type of Change


dsuarezv:
Hi @sarmeyer , would you be so kind to provide an estimation when this will be published to the NPM registry? Thanks!

PR #561, opened 2019-07-25
Avoid the empty div wrapper for 'null' custom controls

luispuig:
### The situation
None element is rendered when a `render*Controls` prop in `null`  but when it is a `function that returns null` the control div wrapper is rendered empty:

image

**Use case**: Removing the previous/next controls when they are not needed returning null depending on `currentSlide` and `slideCount` renders the empty div wrapper causing pointer-events problems.

### The proposal
This PR renders the div wrapper and the element (returned by the `render*Controls` prop) only when the `render*Controls` prop is a function and that function is not returning null/false.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -912,28 +912,32 @@ export default class Carousel extends React.Component {
     } else {
       return this.controlsMap.map(({ funcName, key }) => {
         const func = this.props[funcName];
-        return (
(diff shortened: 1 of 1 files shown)


sarmeyer:
Thank you for submitting this @luispuig!

PR #564, opened 2019-08-08
Export NextButton, PreviousButton, and PagingDots from `main` entrypoint

ahutchings:
This ensures that all exports from src/index.js are exported via the 'main' field.

### Description

This updates the package.json `main` field to point to lib/index.js. Prior to this change, only the default export (Carousel) from src/index.js was available via the `main` entry point, omitting NextButton, PreviousButton, and PagingDots, which are available via the `module` field.

My use case is I want to server-side-render a page containing nuka-carousel, with a reference to PagingDots in order to customize styles.

#### Type of Change

### How Has This Been Tested?

I tested this by making the change in my project where I am using SSR, and this makes it possible to use PagingDots.

In the course of testing this change, I noticed that PreviousButton and NextButton are not in the TypeScript library definition file (index.d.ts). I will submit a separate PR to add those.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,2 +1,0 @@
-module.exports = require('./lib').default;
-module.exports.default = module.exports;
--- a/index.js
+++ b/index.js
@@ -1,0 +1,5 @@
+module.exports = require('./lib').default;
+module.exports.default = module.exports;
+module.exports.NextButton = require('./lib').NextButton;
+module.exports.PreviousButton = require('./lib').PreviousButton;
+module.exports.PagingDots = require('./lib').PagingDots;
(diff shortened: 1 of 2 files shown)


ryan-roemer:
Notes: if our intent is root level exports for outside use (assuming so given content in this PR for motivation) we should re-export everything in root “index.js@

ahutchings:
Thanks @ryan-roemer. I published a reproduction here in case it is useful in testing the proposed fix: https://github.com/ahutchings/nuka-carousel-ssr-components-issue.

sarmeyer:
looks good to me! Thanks for submitting this PR

youngaj:
Sorry for the perhaps stupid question but how I'm still a bit lost on how to override the default styling of the `PagingDots`..  Is there an example I can take a look at?  Seems like the in-line styles would override any settings I would make.  Am I supposed to create my own `PagingDots` component and pass that in instead of the original?

Thanks

Issue #222, opened 2017-06-07
Add rtl support


alvaropinot:
I may have a simple solution to add rtl support just reversing the children and starting the current slide at the end but taking into account how many slides should be shown. Are you accepting PR's ? :D

jonahbgoode:
Any updates on this ?

PR #566, opened 2019-08-09
Feature/rtl

carlomnop:
### Description

Added RTL Checker for the carousel, retracked the issue [https://github.com/FormidableLabs/nuka-carousel/issues/222] #222

#### Type of Change

- [ / ] New feature (non-breaking change which adds functionality)

### How Has This Been Tested?

Tested in the running app (sample app of the package) with a propType of RTL to true.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -970,8 +970,7 @@ export default class Carousel extends React.Component {
       slidesToShow,
       renderAnnounceSlideMessage,
       disableAnimation,
-      rtl,
-      children
+      rtl
     } = this.props;
 
     const duration =
@@ -993,8 +992,7 @@ export default class Carousel extends React.Component {
     const touchEvents = this.getTouchEvents();
     const mouseEvents = this.getMouseEvents();
     const TransitionControl = Transitions[this.props.transitionMode];
-    const validChildren = getValidChildren(this.props.children);
-    const orderedChildren = this.reverseChildren(children, rtl);
+    const validChildren = getValidChildren(this.props.children, rtl);
     return (
       <div
         className={['slider', this.props.className || ''].join(' ')}
@@ -1077,7 +1075,7 @@ export default class Carousel extends React.Component {
                 deltaX={tx}
                 deltaY={ty}
               >
-                {addAccessibility(orderedChildren, slidesToShow, currentSlide)}
+                {addAccessibility(validChildren, slidesToShow, currentSlide)}
               </TransitionControl>
             )}
           />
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -983,7 +994,7 @@ export default class Carousel extends React.Component {
     const mouseEvents = this.getMouseEvents();
     const TransitionControl = Transitions[this.props.transitionMode];
     const validChildren = getValidChildren(this.props.children);
-
+    const orderedChildren = this.reverseChildren(children, rtl);
```
sarmeyer: this should be called with `validChildren` so that you aren't returning any invalid React children from the reverse method

src/index.js
```diff
@@ -1134,7 +1147,8 @@ Carousel.propTypes = {
   withoutControls: PropTypes.bool,
   wrapAround: PropTypes.bool,
   opacityScale: PropTypes.number,
-  slideListMargin: PropTypes.number
+  slideListMargin: PropTypes.number,
```
sarmeyer: would you mind alphabetizing these last two props?

src/index.js
```diff
@@ -1175,7 +1189,8 @@ Carousel.defaultProps = {
   width: '100%',
   withoutControls: false,
   wrapAround: false,
-  slideListMargin: 10
+  slideListMargin: 10,
```
sarmeyer: same comment as above

carlomnop: Thanks for the notice, all changes are in the next push

src/index.js
```diff
@@ -992,6 +1001,7 @@ export default class Carousel extends React.Component {
           getSliderStyles(this.props.width, this.props.height),
           this.props.style
         )}
+        dir="ltr"
```
sarmeyer: where does this `dir` property come in? I don't see it defined anywhere or used anywhere else. if you're trying to use the `rtl` prop then this should be using that prop instead.

src/utilities/bootstrapping-utilities.js
```diff
@@ -32,9 +32,10 @@ export const addAccessibility = (children, slidesToShow, currentSlide) => {
   }
 };
 
-export const getValidChildren = children => {
+export const getValidChildren = (children, rtl) => {
   // .toArray automatically removes invalid React children
-  return React.Children.toArray(children);
+  const childrens = React.Children.toArray(children);
```
sarmeyer: "children" is already plural, can we come up with a better name for this variable since it doesn't matter what the argument is called?

sarmeyer:
It would be great if you could also update the README with this new prop

sarmeyer:
closing. new PR #641 for this feature

PR #576, opened 2019-09-02
Clear all timeouts when component unmounts

sarmeyer:
@namuwikilover I believe the issue you are experiencing is due to the timeouts not being cleared when the component unmounts. While calling `goToSlide` may have fixed your error, I don't believe that it got to the root of the issue. I have made these changes on your branch so you can test it out to see if it also solves your error. If so, I believe this is a better fix as it clears all the timeouts and not just those related to the goToSlide method.


namuwikilover:
@sarmeyer I've tested these commits in my environment. error has not being appeared. Looks good to me too!

PR #580, opened 2019-09-03
Fix a11y issue using aria-hidden attribute (iOS only)

jflayhart:
### Description

In 4.4.2 there were some useful a11y changes to make Nuka more accessible. This was good until after I updated to 4.4.2 and noticed that all our home page carousels on iPhones were NOT accessible anymore.

IMG_1020 TRIM

**Fixes #**

https://github.com/FormidableLabs/nuka-carousel/pull/425/files#r225346008

### How Has This Been Tested?

I made sure there were no regressions in the demo (seems to still be accessible across devices) and our website now behaves as expected on an iOS device with VoiceOver.

### DIFF
--- a/src/utilities/bootstrapping-utilities.js
+++ b/src/utilities/bootstrapping-utilities.js
@@ -1,33 +1,14 @@
 import React from 'react';
 
-export const addAccessibility = (children, slidesToShow, currentSlide) => {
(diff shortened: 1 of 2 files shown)


sarmeyer:
@jflayhart thanks for submitting this fix (and finding it!)

PR #582, opened 2019-09-04
change default prop for height, set to inherit

sarmeyer:
Another effort to get to the bottom of this height/resizing issue! Instead of having the user have to pass in a height prop to the Carousel, or set an explicit height style property on the Carousel or its wrapper, setting the default prop to 'inherit' allows the Carousel to take on the height of its contents. Previously, it was set to 'auto' and it was not inheriting the height value of it's children until a resize event was fired or the slide was changed. 

I tested this by providing a {height: inherit} style property to this codesandbox https://codesandbox.io/s/sad-solomon-gwd10 and did not see the resizing issue. When height was set to auto, the issue persisted.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1179,7 +1179,7 @@ Carousel.defaultProps = {
   edgeEasing: 'easeElasticOut',
   frameOverflow: 'hidden',
   framePadding: '0px',
-  height: 'auto',
+  height: 'inherit',
   heightMode: 'max',
   onResize() {},
   pauseOnHover: true,


YLemelin09:
So the height inherit is by default as in <carousel height='inherit' /> and not  <carousel style={{height:'inherit'}} /> right?

sarmeyer:
@YLemelin09 unless a `height` prop with a different value is assigned to the Carousel, or a height style property is applied to the Carousel manually or by way of a wrapper element. it is essentially the same thing as `<Carousel style={{height:'inherit'}} />`

malcalevak:
@sarmeyer  is https://codesandbox.io/s/sad-solomon-gwd10 not supposed to exhibit the issue? If so, I'd check again in an incognito/private window, as it's still exhibiting the issue. Let me know if I'm completely misunderstanding.

Issue #579, opened 2019-09-03
onDragStart is invoked prematurely when tapping on a slide


jflayhart:
# Bugs and Questions

### Describe Your Environment

- Nuka 4.5.9
- React 16.8.6
- Chrome 76
- Mac OS

### Describe the Problem

`onDragStart` is firing when merely clicking (desktop) or touching (mobile) a slide item. This causes behavior I made within this callback to act prematurely (think, disable vertical scroll) when it should only happen during dragging/swiping behavior.

**Expected behavior:** I expected `onDragStart` to only happen when swiping begins, not a click.

**Actual behavior:** `onDragStart` happens any time a mouse down or touch event occurs on a slide item.

![actual](https://user-images.githubusercontent.com/6808172/64193604-b9178300-ce42-11e9-8de6-f4c8764ff128.gif)

jflayhart:
@sarmeyer Could you please consider this issue and the potential fix (https://github.com/FormidableLabs/nuka-carousel/pull/585)? Trying to rush this out for shop.shipt.com 👍

miguel-square:
> # Bugs and Questions
> ### Describe Your Environment
> * Nuka 4.5.9
> * React 16.8.6
> * Chrome 76
> * Mac OS
> 
> ### Describe the Problem
> `onDragStart` is firing when merely clicking (desktop) or touching (mobile) a slide item. This causes behavior I made within this callback to act prematurely (think, disable vertical scroll) when it should only happen during dragging/swiping behavior.
> 
> **Expected behavior:** I expected `onDragStart` to only happen when swiping begins, not a click.
> 
> **Actual behavior:** `onDragStart` happens any time a mouse down or touch event occurs on a slide item.
> 
> ![actual](https://user-images.githubusercontent.com/6808172/64193604-b9178300-ce42-11e9-8de6-f4c8764ff128.gif) [ ![actual](https://user-images.githubusercontent.com/6808172/64193604-b9178300-ce42-11e9-8de6-f4c8764ff128.gif) ](https://user-images.githubusercontent.com/6808172/64193604-b9178300-ce42-11e9-8de6-f4c8764ff128.gif) [ ](https://user-images.githubusercontent.com/6808172/64193604-b9178300-ce42-11e9-8de6-f4c8764ff128.gif)

Hi, I have this exact problem (That is very well describe). Need to click inside a slide to play a video but it starts dragging in the click before you start swiping or dragging. 

### Enviroment

[...]

PR #585, opened 2019-09-08
Fix onDragStart to happen only when dragging/swiping

jflayhart:
### Description

There is this new prop `onDragStart` that happens `onClick` or `onTouch` rather than an actual drag/swipe event. I modified the code slightly to make the behavior more appropriate to the event. It is better to do it this way so the event doesn't give a false impression that it is starting a drag event when only a click occurs.

Fixes https://github.com/FormidableLabs/nuka-carousel/issues/579

#### Type of Change

### How Has This Been Tested?

Using the demo and writing a unit test.

#### Desktop
desktop

#### Mobile
mobile

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -246,10 +246,6 @@ export default class Carousel extends React.Component {
         };
         this.handleMouseOver();
 
-        if (this.props.onDragStart) {
-          this.props.onDragStart(e);
-        }
-
         this.setState({
           dragging: true
         });
@@ -278,6 +274,11 @@ export default class Carousel extends React.Component {
                 Math.pow(e.touches[0].pageX - this.touchObject.startX, 2)
               )
             );
+
+        if (length >= 10) {
+          if (this.clickDisabled === false) this.props.onDragStart(e);
+          this.clickDisabled = true;
+        }
 
         this.touchObject = {
           startX: this.touchObject.startX,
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -279,6 +275,11 @@ export default class Carousel extends React.Component {
               )
             );
 
+        if (length >= 10) {
+          if (this.clickDisabled === false) this.props.onDragStart(e);
+          this.clickDisabled = true;
```
jflayhart: Probably should have disabled click here anyway, which prevents unwanted click events from firing during swiping on touch devices!

src/index.js
```diff
@@ -1181,6 +1181,7 @@ Carousel.defaultProps = {
   framePadding: '0px',
   height: 'auto',
   heightMode: 'max',
+  onDragStart() {},
```
jflayhart: Added default prop to avoid guards everywhere of `if (this.props.onDragStart) ...`

src/index.js
```diff
@@ -366,7 +363,10 @@ export default class Carousel extends React.Component {
             );
 
         // prevents disabling click just because mouse moves a fraction of a pixel
-        if (length >= 10) this.clickDisabled = true;
+        if (length >= 10) {
+          if (this.clickDisabled === false) this.props.onDragStart(e);
+          this.clickDisabled = true;
```
jflayhart: Don't love coupling these two together, but it works. Any better ideas?

sarmeyer: I can't think of anything right now, we can leave this in until I (or someone else with time) has time to figure out another solution

src/index.js
sarmeyer: thanks!

PR #602, opened 2019-10-15
[performance] Avoid expensive width calculations

jflayhart:
### Description

⚠️ I suggest this be considered as part of a major release around performance improvements. Maybe this can be added to a future release roadmap?

TODO: Other CSS properties like `left`, `margin`, and `height` are also offenders in scroll-transition.js

Under the hood, `scroll-transition.js` animations are doing expensive CSS operations, which cause layout and paint reflows to the browser, which ultimately, where nuka is concerned, affects load time performance. Scripting is also affected as `width` is dynamically computed. This came to our attention at Shipt because it would take ~3 seconds to load our home page! This is because we use many shelves that could potentially contain hundreds of products, so the performance degraded exponentially when one day we had hundreds of products rendered as carousel shelves.

[`width` is expensive](https://csstriggers.com/width) for layout rendering, so I'm trying to cut down on reflows and scripting by making width a static value.

#### Type of Change

- [x] Doc Change remove `slideWidth` prop

⚠️ This will cause a breaking change in that slides are set to `width: auto` and NOT `100%`, meaning most carousels may show slide items differently, so the width must now be set by the user via CSS selectors. **slideWidth** has also been removed to favor performance.

### How Has This Been Tested?

Open demo and make sure everything still works. **To reproduce, you need to add 100+ colors to the nuka demo, and observe load-time behavior.**

Performance Results with 100 shelves:

BEFORE (~3s load time):
before

AFTER  (<2s load time):
after

And that's just one CSS property update that has helped.


src/transitions/scroll-transition.js
```diff
@@ -153,7 +153,7 @@ export default class ScrollTransition extends React.Component {
       height: this.props.vertical
         ? listWidth + spacingOffset
         : this.props.slideHeight,
-      width: this.props.vertical ? 'auto' : listWidth + spacingOffset,
```
jflayhart: Alot of the perf hit was here for non-vertical, scrolling carousels. It would render slides, then set widths accordingly, which causes layout reflows.

jflayhart:
Maybe this can be added to a future release roadmap?

sarmeyer:
@jflayhart Thanks for another contribution to Nuka! Especially in the performance side of things! Could you look into the tests that are failing CI? I'll test it out a bit more, but so far your changes look good. When we get CI passing, we can merge it in - with this and a few other PRs I think we'll have a good release to put out

src/transitions/scroll-transition.js
```diff
@@ -128,7 +128,7 @@ export default class ScrollTransition extends React.Component {
       transform: `scale(${transformScale})`,
       transition: 'transform .4s linear',
       verticalAlign: 'top',
-      width: this.props.vertical ? '100%' : this.props.slideWidth
+      width: this.props.vertical ? '100%' : 'auto'
```
sarmeyer: This doesn't actually work

PR #603, opened 2019-10-16
Fix janky swiping

jflayhart:
## BEFORE
before

## AFTER
after


jflayhart:
oops, meant to do this to my forked version ^_^

PR #604, opened 2019-10-17
Fix issue with webpack builds by updating import

ganemone:
### Description

This fixes a weird issue with the webpack `concatenateModules` optimization config. When this is turned on, the dependency tree is not loaded correctly. I found that using a named import rather than importing Animate by file path corrects this problem. This is the way recommended in the react-move documentation anyway, and modern bundlers (like webpack) can optimize this via tree shaking so there wont be an increase in bundle size.

#### Type of Change

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import ExecutionEnvironment from 'exenv';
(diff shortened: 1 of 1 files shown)


ganemone:
actually this is fixed in the most recent version of webpack

Issue #385, opened 2018-07-23
Vertical Scroll Not Handled in IOS Chrome


DWMiller:
Meaning, if your drag/swipe isn't perfectly horizontal, your scroll will adjust with the vertical drag. 

Haven't managed to confirm why this is, anyone have any insights or able to confirm it's not just something else on my end?

bnhovde:
Having the same issue. This effectively means the user will get stuck on the carousel on all devices with touch.

bnhovde:
Found the culprit. it's the `touch-action: none` style applied to the `<ul class="slider-list">...</ul>`

A temporary hack:
```
  .slider-list {
    touch-action: initial !important;
  }
```

DWMiller:
hmm, I don't see a touch-action: none on that element, but touch-action: pan-y pinch-zoom.

bnhovde:
@DWMiller - Hmm, I was on version `4.2.5`. I updated to `4.3.5` and that resolved the above. What version are you on?

@ryan-roemer PS: After updating to the latest version the above issue seems to be resolved.

DWMiller:
@bnhovde `4.3.2` here,  I'll give it a bump and check back.

sarmeyer:
@DWMiller did upgrading fix the issue you were seeing?

DWMiller:
Sorry for the lack of follow up here, and no, this issue was not resolved. 

I'm also able to reproduce the problem on http://kenwheeler.github.io/nuka-carousel. I'm aware that that is an older version, but same issue.

Anyone have any links to sites using this I could compare to? (I'd also test the demo in the repo itself, but installing it on windows seems to be a no go)

ScottSmith95:
I can confirm that @bnhovde's solution works. This is related to https://github.com/FormidableLabs/nuka-carousel/issues/447, in my opinion.

missmimia:
> @DWMiller - Hmm, I was on version `4.2.5`. I updated to `4.3.5` and that resolved the above. What version are you on?
> 
> @ryan-roemer PS: After updating to the latest version the above issue seems to be resolved.

meet the same thing...

set the style you said, but it doesn't work.

my version is "nuka-carousel": "^4.4.3", IOS safari.

sarmeyer:
fixed in #609

PR #609, opened 2019-11-07
prevent scroll when dragging on ios

Lopliok:
### Description

This issue happens on ios in safari. When user slides carousel, it could scroll page but it shouldn't be like that. 

Fixes # (issue)
#385

#### Type of Change

### How Has This Been Tested?

Tested on Iphone 5s
ios 11


Lopliok:
@sarmeyer

PR #633, opened 2020-01-07
Issue 618: Fixed max (and first) height mode calculations

### LINKED ISSUE #618: heightMode 'max' is not working
v4.5.13

When heightMode is set to 'max'  ul.slider-list receives correct inline height property value, but all of the slides have height set to 'auto', which pretty much blows all the things away. I can't use min-height: 100% on the slider-slide anyways, because the container which is inside of the slider-slide can't inherit the height neither from slider-slide nor from slider-list

mariano-formidable:
### Description

This PR updates the way we calculate a slide's height in `max` or `first` mode, so we can stamp each slide's containing `li` with a specific height in pixels, instead of using `auto`.

The update has been applied to `scroll-transition`, `fade-transition`, and `3d-scroll-transition`.

Fixes #618 

#### Type of Change

### How Has This Been Tested?

I've updated our bootstraping-utilities tests to validate the new height calculations based off a slide's children.

### DIFF
--- a/src/transitions/scroll-transition.js
+++ b/src/transitions/scroll-transition.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { getSlideHeight } from '../utilities/style-utilities';
 
 const MIN_ZOOM_SCALE = 0;
 const MAX_ZOOM_SCALE = 1;
@@ -140,16 +141,17 @@ export default class ScrollTransition extends React.Component {
       }
     }
 
-    return targetPosition + offset;
+    return targetPosition + offset || 0;
   }
   /* eslint-enable complexity */
-
   formatChildren(children) {
     const { top, left, currentSlide, slidesToShow } = this.props;
     const positionValue = this.props.vertical ? top : left;
+
     return React.Children.map(children, (child, index) => {
       const visible =
         index >= currentSlide && index < currentSlide + slidesToShow;
+
       return (
         <li
           className={`slider-slide${visible ? ' slide-visible' : ''}`}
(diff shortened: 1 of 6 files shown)


src/transitions/scroll-transition.js
```diff
@@ -140,16 +140,17 @@ export default class ScrollTransition extends React.Component {
       }
     }
 
-    return targetPosition + offset;
+    return targetPosition + offset || 0;
```
mariano-formidable: Added a default return value of 0.  It looks like in some of our tests (with things not fully configured) it was returning `NaN` which was giving us a warning.  I figure if it calculates to `NaN` we can just send back 0 so things don't break (and again, it seems this only occurred in our tests).

src/transitions/scroll-transition.js
```diff
@@ -171,10 +173,21 @@ export default class ScrollTransition extends React.Component {
...
+      : this.props.slideHeight;
+
     return {
       boxSizing: 'border-box',
       display: this.props.vertical ? 'block' : 'inline-block',
-      height: 'auto',
+      height:
+        calculatedHeight > 0 && this.props.heightMode !== 'current'
+          ? calculatedHeight
+          : 'auto',
```
mariano-formidable: When the carousel first loads the calculatedHeight will be 0, so we need to let it render with `auto` at first so the heights can be correctly registered and then we swap out the `auto` to a fixed height.

test/specs/bootstrapping-utilities.test.js
```diff
@@ -4,44 +4,89 @@ import {
...
+        }, // 3
+        {
+          offsetHeight: 0,
+          children: [
+            { offsetHeight: 100 },
+            { offsetHeight: 100 },
+            { offsetHeight: 300 }
+          ]
+        } // 4
+      ];
```
mariano-formidable: To be able to use the same shared tests, I made sure the children elements totaled up to the same `offsetHeight` as the `noChildrenSlides` slides.

PR #635, opened 2020-01-08
Issue 494: Updated scroll transition logic to fix wrapAround flash

### LINKED ISSUE #494: Flash after wrap around
- nuka-carousel: ^4.4.7
- react: ^16.7.0, react-dom: ^16.7.0
- Chrome Version 71.0.3578.98
- Surface Book 2

##Problem

https://codesandbox.io/s/3r8110mwlp

**Expected behavior:** Carousel goes slide 1 ,slide 2 , slide 3, slide 1 seamlessly 
**Actual behavior:** After slide 3, carousel returns to slide 1 but flashes 

### Additional Information
Looks like slide 1 goes to -4500px and jumps to 0px when performing the wrap around

mariano-formidable:
### Description

This PR updates the `scroll-transition` logic to remove the wrapAround flash.  It divides up the available slides to be in-front and in-back of the carousel (taking into account, alignment _and_ slides to show) so flipping through slides shouldn't flash (as much...).

Fixes #494 

#### Type of Change

### How Has This Been Tested?

Passes all written tests, and lots of manual testing.

### Screenshots

Normal wrap around: 
normal-wraparound

Partial wrap around:
partial-wraparound

Height Mode "current" wrap around:
current-wraparound

Extreme wrap around.  Slides to show is 6, slides to scroll is 2, and there are _only_ 6 slides in total :scream: (still manages to "work" as best it can):
extreme-wraparound

### DIFF
--- a/demo/app.js
+++ b/demo/app.js
@@ -15,6 +15,7 @@ class App extends React.Component {
       underlineHeader: false,
       zoomScale: 0.5,
       slidesToShow: 1,
+      slidesToScroll: 1,
       cellAlign: 'left',
       transitionMode: 'scroll',
       heightMode: 'max',
--- a/src/index.js
+++ b/src/index.js
@@ -17,6 +17,7 @@ import {
   calcSomeInitialState
 } from './utilities/utilities';
 import {
+  getAlignmentOffset,
   getImgTagStyles,
   getDecoratorStyles,
   getSliderStyles,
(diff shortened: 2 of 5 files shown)


src/index.js
```diff
@@ -974,13 +958,10 @@ export default class Carousel extends React.Component {
...
         slidesToShow,
         slideWidth,
-        cellAlign,
-        left: props.vertical ? 0 : this.getTargetLeft(),
-        top: props.vertical ? this.getTargetLeft() : 0
+        cellAlign
       },
       () => {
         stateCb();
-        this.setLeft();
```
mariano-formidable: Removing the call to `setLeft` and the updates to `left` and `top` from the method `setDimensions`, fixed the wrap around problem we saw when `heightMode` was set to `current`.  I beliiieeeeve (but I'm not positive), that these extra calls to set the `left` and `top`, was just paranoiac ways to ensure that the carousel wouldn't blink.  But, I think the new wrapAround logic should prevent it sooooooooo... I don't think we need these calls.  I've been playing around with the demo a bit and it all still works.

demo/app.js
```diff
@@ -15,6 +15,7 @@ class App extends React.Component {
       underlineHeader: false,
       zoomScale: 0.5,
       slidesToShow: 1,
+      slidesToScroll: 1,
```
mariano-formidable: Added this to state in our demo app so we can easier update this param when testing.

PR #639, opened 2020-01-15
Issue 623: Keyboard controls will only work when nuka is in focus.

### LINKED ISSUE #623: Carousel keyboard controls trigger on all carousels
# Bugs and Questions

**Expected behavior:** Press left or right when the focus is on a certain carousel and only this carousel behaves to the command
**Actual behavior:** All the carousels in the page scroll together

### Additional Information
I have a netflix like page with multiple carousels on the same page. If I tab to focus in a carousel and use the keyboard controls, all the carousels get the scroll trigger instead of just the one I'm focused in.

Maybe carousels could have a key prop and make sure to have a tab index just on the slider div and not the children.

mariano-formidable:
### Description

This PR updates our keyboard controls so they'll only work if `enableKeyboardControls` is `true` _and_ the carousel is in focus.

**Note:**
Technically, this introduces a breaking change because _before_ the keyboard controls worked even if the carousel wasn't in focus (although, not sure how many people were actually taking advantage of that or appreciated that fact).  If we want, we can add in an additional prop that'll let the keyboard controls work regardless of carousel focus.  

Fixes #623 

#### Type of Change

### How Has This Been Tested?

Manually using the demo, and fixed existing tests as needed so they pass.

## Screenshots

In the animation I click on the top carousel and the navigate with keyboard, and then click on bottom one and animate with the keyboard again:
focus-controls


src/transitions/3d-scroll-transition.js
```diff
@@ -60,6 +61,8 @@ export default class ScrollTransition3D extends React.Component {
           }`}
           style={this.getSlideStyles(index, positionValue)}
           key={index}
+          onClick={handleSelfFocus}
+          tabIndex={-1}
```
mariano-formidable: Setting to `-1` lets us programmatically set the focus without having to set any specific tab index.

YouKnowEno:
I have a use case in which I would like to enable keyboard controls even if the carousel is not in focus. Is it possible to modify enableKeyboardControls to enable this use case, if so desired? Thank you!

jkrehm:
Ditto, I'm also trying to figure out a work-around. I wish this had been an opt-in change, i.e. consuming application tracks the focus and toggles keyboard controls on/off. I think it should have been called out as a breaking change with a major bump.

TomGeshury:
Hello a workaround I found was exposing the nuka carousel ref in the component its being called in and in ComponentDidMount focusing the ref.

So it looks like this this:

```
componentDidMount() {
// (the stateful logic that handles focusing the carousel)
this.carouselRef.handleFocus()
}
<Carousel
          ref={(carouselRef) => {
                this.carouselRef = carouselRef;
              }}
/>
```

PR #641, opened 2020-01-20
Feature/rtl

mm22dev:
### Description

This PR intends to continue the work proposed in https://github.com/FormidableLabs/nuka-carousel/pull/566 with the update of the README and the assignment of the dir property through the prop rtl.

When the page is displayed in right-to-left mode - for instance applying the `dir` attribute set to `rtl` to the html tag - that attribute spreads over all descendants. 
By forcing the `dir` attribute of the `.slider` wrapper to `ltr` we make sure that `.slider-frame` and the `.slider-control-bottomcenter` are not affected by strange behaviour.

Fixes # (issue)

### How Has This Been Tested?

Tested in the demo app


package.json
```diff
@@ -91,7 +91,7 @@
   },
   "repository": {
     "type": "git",
-    "url": "https://github.com/kenwheeler/nuka-carousel.git"
+    "url": "https://github.com/carlomnop/nuka-carousel.git"
```
sarmeyer: why were these changed?

src/index.js
```diff
@@ -1073,6 +1082,7 @@ export default class Carousel extends React.Component {
           getSliderStyles(this.props.width, this.props.height),
           this.props.style
         )}
+        dir={rtl ? 'ltr' : 'auto'}
```
sarmeyer: where does `ltr` come in? and `auto`? it doesn't look like `dir` is being used anywhere

.gitignore
```diff
@@ -30,8 +30,5 @@ $RECYCLE.BIN/
 node_modules/
 bower_components/
 .tmp
-dist
```
sarmeyer: is there a reason these needed to be removed?

package.json
mm22dev: My bad, I forgot to update. I started working from the fork of the repo https://github.com/carlomnop/nuka-carousel/tree/master from which these changes derive.

src/index.js
mm22dev: As you well say, `dir` attribute is not used anywhere.
According to https://www.w3.org/International/questions/qa-html-dir 

> In HTML the base direction is either set explicitly by the nearest parent element that uses the dir attribute, or, in the absence of such an attribute, the base direction is inherited from the default direction of the document, which is left-to-right (LTR).

This means that the application of `dir="rtl"` to an ancestor element of the `<Carousel />` is inherited from the `<Carousel />`, causing a wrong rendering of `.slider-list` and its controls.

The proposed modification is intended to overwrite the `dir` attribute of the `.slider` element assuming that if a flipped version is required (providing the prop `rtl` to the `<Carousel />`) an ancestor element could have set the attribute `dir="ltr"`.

.gitignore
mm22dev: Same as comment on top about package.json.

index.d.ts
```diff
@@ -370,6 +376,13 @@ export interface CarouselProps {
    * @default false
    */
   withoutControls?: boolean;
+
+  /**
+   * Sets the RTL to default
```
sarmeyer: This description doesn't really make sense. The type doesn't set the `rtl` to default, this prop is used to set the direction of the slides. So this should read something more like that

src/index.js
```diff
@@ -1047,6 +1050,10 @@ export default class Carousel extends React.Component {
     }
   }
 
+  reverseChildren(children, rtl) {
+    return rtl ? [...children].reverse() : children;
```
sarmeyer: this function should only return reversed children because it's called `reverseChildren`. It's confusing that it can return `children` also.

sarmeyer: Also upon further review, I don't see this method being called anywhere

PR #642, opened 2020-01-20
Issue 622: Slides that are not fully visible cannot receive focus.

mariano-formidable:
### Description

This PR adds a package called `wicg-inert`(https://www.npmjs.com/package/wicg-inert), which implements a polyfill for the html attribute `inert`.  Using the polyfill, when an html element is tagged as `inert`, the element (and its children) is removed from the tab index.  This PR tags every non-fully visible slide as `insert`, so when tabbing through a page with a nuka carousel you will _not_ tab into hidden slides.

Fixes #622 

#### Type of Change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)

### How Has This Been Tested?

It's been manually tested with a few configurations, and all existing tests pass.

### Screenshots

The below screenshots show a subtle blue outline on elements as they receive tab focus, notice that when the input box within a slide receives focus, the next element to receive focus is either the input box in another _fully_ visible slide or it jumps to the prev/next buttons.  It no longer jumps focus to slides that are _not_ visible.

Normal slides tab focus:
standard-tabfocus

Partial slides tab focus (with alignment):
partial-tabfocus

Extreme slides tab focus (focus moves to both visible slides, even on wrap around):
extreme-tabfocus

PR #643, opened 2020-01-21
Issue 622: Slides that are not fully visible cannot receive focus.

### LINKED ISSUE #622: Using tab to scroll pushes the slider to the left
# Bugs and Questions

### Prerequisites

- [x] I've searched open [issues](https://github.com/FormidableLabs/nuka-carousel/issues) to make sure I'm not opening a duplicate issue

### Describe Your Environment

- What version of nuka-carousel are you using?  4.5.13
- What version of React are you using? 16.9.0
- What browser are you using? Chrome
- What machine are you on? Macbook pro 15

### Describe the Problem

I cannot share my code but I dont think you will need it to understand the problem.

With a carousel showing multiple items, i use tab to check the accessibility in my page. It tabs on all slides and when it passes from the last visible slide to the ones that are hidden in the overflow, the .slider div is pushed on the left to show the hidden slides thus pushing the centerRightControl in the middle of the page where the last shown slide should be.

**Expected behavior:** Either when i tab the whole carousel is selected and the next tab skips to the DOM element after the carousel or I can tab only on the visible slides

**Actual behavior:** what i said earlier

### Additional Information

Expected:

carrousel_working_1

Reality: 

carrousel_notworking_2

mariano-formidable:
### Description

This PR is the same as this PR (https://github.com/FormidableLabs/nuka-carousel/pull/642), it's just that the merge target is master.

### DIFF
--- a/src/utilities/utilities.js
+++ b/src/utilities/utilities.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import { getAlignmentOffset } from './style-utilities';
 
 export const addEvent = function(elem, type, eventHandle) {
   if (elem === null || typeof elem === 'undefined') {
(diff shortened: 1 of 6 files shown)


src/utilities/utilities.js
```diff
@@ -169,3 +170,44 @@ export const handleSelfFocus = e => {
...
+    currentSlide,
+    cellSpacing,
+    slideCount,
+    slideWidth,
+    frameWidth,
+    wrapAround
+  } = config;
+
+  // Slide width can't be 0
+  const fullSlideWidth = slideWidth + cellSpacing || 1;
```
mariano-formidable: Added this so tests would pass because most of our tests don't actually set a slideWidth (because the virtual dom doesn't have one).

src/utilities/utilities.js
```diff
@@ -169,3 +170,44 @@ export const handleSelfFocus = e => {
     e.currentTarget.focus();
   }
 };
+
+export const isFullyVisible = (slideIndex, config) => {
```
mariano-formidable: I also removed `positionValue` from the passed in arguments since I realized I wasn't even using it :scream:

YLemelin09:
Will this be in a release of it is usable as soon as you merge on master? I have Updated to 4.6.1 and I still can tab passed the visible slides and the it still pushes the right control to the middle of the view like in https://github.com/FormidableLabs/nuka-carousel/issues/622

PR #644, opened 2020-01-22
Issue 611: Added extra props to allow customization of default controls

### LINKED ISSUE #611: Next/Prev-Btn & PagingDots not configurable
It's very easy to make them via props configurable.

E.g. color of paging dots, next/prev btn text/icon, etc.

But if you just want to make a fine configuration, like changing the color of the paging dots, then creating a whole component/extending the old one is a bit overkill and decreases code quality/readability.

As this feature is really simple to implement via props, I would love to see it in a future release.

See #607

mariano-formidable:
### Description

This PR adds extra props to let users customize Nuka's default controls.

**NOTE:**  I did _not_ implement the ability to swap out the text of the Next/Prev button for icons or components.  A custom icon would also need to support hover effects, enable/disable, and maybe other functionality that, I felt, could introduce too many edge cases.  If a user wants totally special Next/Prev buttons, then, I feel, it's best to simply pass in their own custom component and handle all the logic themselves.

Fixes #611 

#### Type of Change

### How Has This Been Tested?

Manually tested, and current tests pass.

### Screenshots

Normal demo app, without any changes (meaning things keep working by default):
normal-custom

Shnazzy demo app with custom text for Prev/Next buttons and red paging dots:
custom-custom

### DIFF
--- a/src/default-controls.js
+++ b/src/default-controls.js
@@ -16,23 +16,36 @@ export class PreviousButton extends React.Component {
     super(...arguments);
     this.handleClick = this.handleClick.bind(this);
   }
+
   handleClick(event) {
     event.preventDefault();
     this.props.previousSlide();
   }
+
   render() {
+    const {
+      prevButtonClassName,
+      prevButtonStyle = {},
+      prevButtonText
+    } = this.props.defaultControlsConfig;
+
     const disabled =
       (this.props.currentSlide === 0 && !this.props.wrapAround) ||
       this.props.slideCount === 0;
+
     return (
       <button
-        style={defaultButtonStyles(disabled)}
+        className={prevButtonClassName}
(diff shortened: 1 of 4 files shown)


src/default-controls.js
```diff
@@ -181,33 +205,35 @@ export class PagingDots extends React.Component {
...
+                }}
                 onClick={this.props.goToSlide.bind(null, index)}
                 aria-label={`slide ${index + 1} bullet`}
               >
                 <svg className="paging-dot" width="6" height="6">
-                  <circle
-                    cx="3"
-                    cy="3"
-                    r="3"
-                    style={{
```
sarmeyer: Yes, thank you for moving this to the buttonStyles!

PR #659, opened 2020-01-30
fix(inert polyfill): lazy load the inert polyfill to fix issues when SSR

### LINKED ISSUE #658: ReferenceError: Element is not defined
The newest version 4.6.2 includes an error referring `wicg-inert`: 

image

#622 #642 #643

matthaywardwebdesign:
### Description

When using nuka-carousel in an environment which uses server side rendering (SSR). The loading of the wicg-inert polyfill causes the build to fail due to Element not being present when SSR.

Fixes #658 

#### Type of Change

### How Has This Been Tested?

- Tested both SSR and non SSR to ensure working as expected.
- Confirmed the polyfill was being loaded when the app was being hydrated on the client side.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,5 +1,4 @@
 import React from 'react';
-import 'wicg-inert';
 import PropTypes from 'prop-types';
 import ExecutionEnvironment from 'exenv';
 import Animate from 'react-move/Animate';
@@ -110,6 +109,9 @@ export default class Carousel extends React.Component {
   componentDidMount() {
     // see https://github.com/facebook/react/issues/3417#issuecomment-121649937
     this.mounted = true;
+
+    // Polyfills are dynamically loaded in componentDidMount to fix issues when SSR
+    this.loadPolyfills();
     this.setLeft();
     this.setDimensions();
     this.bindEvents();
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -236,6 +238,12 @@ export default class Carousel extends React.Component {
     this.getlockScrollEvents().unlockTouchScroll();
   }
 
+  loadPolyfills() {
+    if (typeof Element !== 'undefined') {
+      import('wicg-inert');
```
JoviDeCroock: lazy-loading isn’t supported in libraries we need to handle this on the app since we need a specific operation your bundler performs. In library code this won’t happen and will cause issues in some browsers. We need to fix this in `wicg-inert` itself

This will most likely work in some setups like next.js and modern browsers but not in all

matthaywardwebdesign: Ah very good point! Thanks! I've raised a PR in wicg-inert :)

yangxuanxing:
i have the same problem, wicg-inert fails when ssr;
can we just recommend to use wicg-inert in projects, but not import it in this library?

sarmeyer:
I'm closing this since the PR update to wicg-inert has been released with v3.0.3 and updated in Nuka in the latest release (coming today) through #667

PR #670, opened 2020-02-27
Feature 637: Added scrollMode prop

### LINKED ISSUE #637: ScrollMode prop
## Enhancement Request

Some users may want the option to toggle whether the empty slides are shown at the end of the carousel when slidesToScroll is greater than the remaining number of slides. This feature was part of the carousel at one point in time, but I think it was lost or reverted during a refactor. This will not be a high priority fix, unless someone has a compelling case for why it should be. I just want to make sure the request from about a year ago is not lost. 

Original Issue: https://github.com/FormidableLabs/nuka-carousel/issues/13
Original PR: https://github.com/FormidableLabs/nuka-carousel/pull/155

```
      // If scrollMode = remainder, only scroll the amount of slides necessary without showing blank slides.
      if (this.props.scrollMode === 'remainder') {
        this.goToSlide(Math.min(this.state.currentSlide + this.state.slidesToScroll, childrenCount - slidesToShow));
      } else if (this.props.scrollMode === 'page') {
        // Otherwise, slidesToScroll always equals slides to show
        this.goToSlide(this.state.currentSlide + this.state.slidesToScroll);
      }
```

```
Usage: When there are 10 slides and slidesToScroll = 4 & slidesToShow = 4, then clicking "next" will result in items 5-8 showing, clicking "next" again will have two different modes to handle this.

If scrollMode = 'remainder', then slides 7-10 will show to avoid blank spaces in the slider view.

If scrollMode = 'page', then slides 9-10 will show with 2 remaining empty slide slots.

scrollMode defaults to 'remainder' to avoid breaking existing usage.

```

mariano-formidable:
This PR adds the `scrollMode` prop to `nuka`.  When it's set to `page`, nuka will set `slidesToScroll` equal to `slidesToShow` and page through each slide which _can_ result in blank slides at the end.  However, if `scrollMode` is set to `remainder`, then `nuka` will behave at it normally does.

Fixes #637 

#### Type of Change

### How Has This Been Tested?

Manually tested, and all the CI tests pass.

### Screenshots

Scroll mode set to `remainder`, `slidesToShow=2`, `slidesToScroll=1`: 
remainder-v2

Scroll mode set to `page`, `slidesToShow=2`, nuka ignores the prop `slidesToShow':
page


sarmeyer:
I think we just need to define a type for ScrollMode in index.d.ts - otherwise looks good to me

mariano-formidable:
> I think we just need to define a type for ScrollMode in index.d.ts - otherwise looks good to me

Ohhhhhh yeah, I forgot nuka now has internal types.  I'll update the type file and let you know when you can re-review.

index.d.ts
```diff
@@ -341,6 +343,12 @@ export interface CarouselProps {
     slideCount
   }: CarouselSlideRenderControlProps) => string;
 
+  /**
+   * Supports 'page' and 'remainder' scroll modes.
+   * Defaults to 'remainder'.
+   */
+  scrollMode?: CarouselScrollModeProp;
```
mariano-formidable: Updated the type defs 👍

PR #673, opened 2020-03-04
Fix: Removed use of deprecated UNSAFE_componentWillReceiveProps

mariano-formidable:
### Description

This PR removes the use of the deprecated `UNSAFE_componentWillReceiveProps` event listener.  The logic within the deprecated method was moved to `componentDidUpdate` and reworked to work with previous props instead of next props.

#### Type of Change

### How Has This Been Tested?

Lots of manual testing, all existing tests pass (which at first they didn't until I adjusted the code).

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -25,6 +25,7 @@
     "react/prop-types": 0,
     "react/sort-prop-types": 0,
     "react/no-string-refs": 0,
+    "react/no-did-update-set-state": 0,
     "max-statements": 0,
     "no-invalid-this": 0
   }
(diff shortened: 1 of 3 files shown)


.eslintrc
```diff
@@ -25,6 +25,7 @@
     "react/prop-types": 0,
     "react/sort-prop-types": 0,
     "react/no-string-refs": 0,
+    "react/no-did-update-set-state": 0,
```
mariano-formidable: Needed this so I could move the componentWillReceiveProps methods into componentDidUpdate.

test/specs/carousel.test.js
```diff
@@ -363,6 +363,7 @@ describe('<Carousel />', () => {
       expect(wrapper).toHaveState('currentSlide', 2);
       wrapper.setProps({ children: [] });
       expect(wrapper).toHaveState('currentSlide', 0);
+      wrapper.update();
```
mariano-formidable: This test suddenly started failing with the change unless I triggered a wrapper update to sync the latest render updates.

sarmeyer:
looks good!

PR #674, opened 2020-03-05
Chore: Upgraded react-move package to reduce overall bundle size

mariano-formidable:
### Description

I used `webpack-bundle-analyzer` to analyze our bundle size.  Fortunately (or unfortunately), it doesn't look like we have any extra size to trim.  The only thing I was able to do to reduce bundle size was upgrade `react-move` to `6.x`.  This reduced the size by a little because `5.x` has extra code to support older versions of react, but the reduction was only about 2 kb 😱 

Our bundle size has grown simply because we support so many features.  If we wanted to really reduce size, we'd have to either re-write large chunks of code to do the same with less lines of code (might be possible in some areas), or maybe separate out the transition scripts from the core bundle and require users to also include the specific transition scripts they want to use 🤔 

### How Has This Been Tested?

Manually and passes all the tests.

### Screenshots

Before the update to `react-move` the parsed size was 71.3 KB, and now it's... 69.4 KB 🎉 
Screen Shot 2020-03-05 at 11 46 52 AM

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "d3-ease": "^1.0.3",
     "exenv": "^1.2.0",
     "prop-types": "^15.6.0",
-    "react-move": "^5.0.0"
+    "react-move": "^6.1.0"
   },
   "devDependencies": {
     "@babel/cli": "^7.5.5",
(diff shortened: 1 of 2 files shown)


sarmeyer:
I might be in favor of separating the transition scripts from the core bundle, since people may only use one type of transition, but this is good for now

PR #706, opened 2020-07-07
Add a more descritive info about wrapAround option

### LINKED ISSUE #8: Loop to first/last
Is there anyway to loop slides ? Like when on the last slide onSwype redirect to first.

lucasfloriani:
### Description

Add a more descriptive info about wrapAround option that is similar from other carousel libs options (repeat or infinite).
I only finded the option after a good time searching how to do it, principally from the fact that I remember this option to be used with the "keyword" repeat or infinite like the comments in Issue #8 .
With this change in the docs, people that will eventually find the repo will have a more easy way to find/understand this option.

Fixes #8 

#### Type of Change

### How Has This Been Tested?

Only markdown changes, not changed real code.

OBS: Sorry for my bad english, not my first language.


sarmeyer:
Thank you for this!

Issue #741, opened 2020-11-26
ARIA props shouldn't take precedence over the component's child props


halitogunc:
### Describe Your Environment

- What version of nuka-carousel are you using?
4.7.3
- What version of React are you using?
16.14.0
- What browser are you using?
Version 86.0.4240.198 (Official Build) (x86_64)
- What machine are you on?
macOS 10.15.7 (19H2)

### Describe the Problem
Aria props are always overriding component's child props for `aria-hidden` and `tabindex` attributes. Child component's props should be able to change these values if defined.

PR #742, opened 2020-11-26
fix: child.props should take precedence over aria props

halitogunc:
This PR fixes an issue that was created in #741 

#### Type of Change


sarmeyer (approved):

sarmeyer:
@halitogunc thanks for submitting this PR!

PR #791, opened 2021-07-07
fix: bind blockEvent method to this

### LINKED ISSUE #787: Memory Leak - touchmove event listeners
### Describe Your Environment

- What version of nuka-carousel are you using? **4.7.8**
- What version of React are you using? **16.10**
- What browser are you using? **Chrome**

### Describe the Problem

Then handling of `touchmove` event listeners produces a memory leak as they are never cleaned up.

```
getLockScrollEvents() {
    const blockEvent = (e) => {
      if (this.state.dragging) {
        const direction = swipeDirection(
          this.touchObject.startX,
          e.touches[0].pageX,
          this.touchObject.startY,
          e.touches[0].pageY,
          this.props.vertical
        );
        if (direction !== 0) {
          e.preventDefault();
        }
      }
    };

    const lockTouchScroll = () => {
      document.addEventListener('touchmove', blockEvent, {
        passive: false
      });
    };

    const unlockTouchScroll = () => {
      document.removeEventListener('touchmove', blockEvent, {
        passive: false
      });
    };

    return {
      lockTouchScroll,
      unlockTouchScroll
    };
  }
```

Because the `blockEvent` function is recreated every time the method is run, the `unlockTouchScroll` cleanup that happens through `componentWillUnmount` never removes the original handler.  `blockEvent` should probably be bound to `this` (like the other event handlers) so it can be properly removed.

CleanShot 2021-07-02 at 11 41 09@2x

*^^ After repeated mount/unmount cycles of the carousel component*

### Additional Information
Any additional information, configuration or data that might be necessary to reproduce the issue.

sarmeyer:
### Description

Bind blockEvent method to `this` in order to properly remove the touchmove event listener on unmount

Fixes #787 

#### Type of Change

### How Has This Been Tested?

`npm run test`
manual testing

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -83,6 +83,7 @@ export default class Carousel extends React.Component {
     this.getOffsetDeltas = this.getOffsetDeltas.bind(this);
     this.getTargetLeft = this.getTargetLeft.bind(this);
     this.getTouchEvents = this.getTouchEvents.bind(this);
+    this.blockEvent = this.blockEvent.bind(this);
     this.goToSlide = this.goToSlide.bind(this);
     this.handleFocus = this.handleFocus.bind(this);
     this.handleBlur = this.handleBlur.bind(this);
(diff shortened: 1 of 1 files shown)


markphilpot:
This looks right! 👍

PR #796, opened 2021-08-18
Perf: Use more modern screen-reader only styles

kurtextrem:
### Description

See: https://a11y-guidelines.orange.com/en/web/components-examples/accessible-hiding/. 

This improves performance, since the previously used `left: -1000px` technique created huge layers when the div was close to (or overlapped) an element which got promoted to its own layer.

#### Type of Change

### How Has This Been Tested?

See screenshot: https://prnt.sc/1qae4pk
After swapping the css, the layers are gone.

### DIFF
--- a/src/announce-slide.js
+++ b/src/announce-slide.js
@@ -3,11 +3,14 @@ import React from 'react';
 const AnnounceSlide = ({ message }) => {
   const styles = {
     position: 'absolute',
-    left: '-10000px',
-    top: 'auto',
     width: '1px',
     height: '1px',
-    overflow: 'hidden'
+    overflow: 'hidden',
+    padding: 0,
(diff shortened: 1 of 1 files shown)


src/announce-slide.js
```diff
@@ -3,11 +3,14 @@ import React from 'react';
...
     position: 'absolute',
-    left: '-10000px',
-    top: 'auto',
     width: '1px',
     height: '1px',
-    overflow: 'hidden'
+    overflow: 'hidden',
+    padding: 0,
+    margin: '-1px',
+    clip: 'rect(0, 0, 0, 0)',
```
XavierMeunier, line 11: `clip` is going to be deprecated and `clip-path` seems to be the new way to go.

PR #802, opened 2021-09-20
fix: small a11y semantic improvements

### LINKED ISSUE #681: How is accessibility supported with screen readers?
# Bugs and Questions

### Prerequisites

- [x] I've searched open [issues](https://github.com/FormidableLabs/nuka-carousel/issues) to make sure I'm not opening a duplicate issue

### Describe Your Environment

- What version of nuka-carousel are you using? **4.6.6**
- What version of React are you using? **16.11.0**
- What browser are you using? **Chrome latest**
- What machine are you on? **Mac/Win7**

### Describe the Problem

Apologies if this is more a question / how to, rather than an actual issue.

What I'm trying to figure out is how to make the carousel work nicely with screen readers and NVDA in particular.

**My scenario:** I have a carousel with multiple slides (say 4 slides, as an example)

**Expected behavior:** When the carousel is focused, the screen reader only reads out the visible slide. The screen reader will read out subsequent slides as and when they appear on the screen, after either user interaction or autoplay.

**Actual behavior:** When the carousel is focused, the screen reader reads out all the slides in one go, the one after the other (regardless of which slide is actually displayed). When a slide changes, either though user interaction or autoplay, the screen reader remains silent.

### Additional Information

Having inspected the DOM, it looks like all four slides are present in DOM when the carousel loads, regardless of which one is currently visible. So, technically the screen reader treats the whole slide set as a `<ul/>` with four <`li/>` children and reads the whole thing out. There seem to be no `aria-*` or `role` attributes that will help the reader know which slide is currently visible and which is not.

Any ideas on how the desired behavior can be achieved will be greatly appreciated! Hopefully I'm just missing something in the way I'm using nuka-carousel :) 

### Update after a bit of investigation 
I had a look at the code and realised the following:
- In `src/index` the `addAccessibility` method from `./utilities/bootstrapping-utilities` is imported. However, this method does nothing meaningful, as both routes of the conditional within do the same (no)thing. It looks like the `addAccessibility` method from `./utilities/utilities` should have been imported instead?
- Even so, this would only apply when more than ones slides are being shown at a time with `slidesToShoe` variable
- Again, even if we overcome the previous obstacle, the actual `aria-hidden` attribute is eventually being ignored by the renderer of the transition e.g. `scroll-transition`

So I think that the current version of the carousel is not really supporting screen readers with the a`aria-hidden` attribute. Whether this is intentional or I not I cannot say (maybe people prefer to have the screen reader read out all slides, even the hidden ones

sarmeyer:
### Description

The carousel needs some improvements to its accessibility and use with screen readers

Fixes #681

#### Type of Change

### How Has This Been Tested?

Manually and with unit tests

### DIFF
--- a/demo/app.js
+++ b/demo/app.js
@@ -86,7 +86,7 @@ export default function App() {
         wrapAround={wrapAround}
         zoomScale={Number(zoomScale || 0)}
         renderAnnounceSlideMessage={({ currentSlide, slideCount }) => {
-          return `Showing slide ${currentSlide + 1} of ${slideCount}`;
+          return `Showing slides ${currentSlide + 1} of ${slideCount}`;
         }}
         renderTopCenterControls={({ currentSlide }) =>
           renderTopControls(currentSlide)
--- a/src/announce-slide.js
+++ b/src/announce-slide.js
@@ -9,7 +9,7 @@ const AnnounceSlide = ({ message }) => {
     padding: 0,
     margin: '-1px',
     clip: 'rect(0, 0, 0, 0)',
-    'white-space': 'nowrap',
+    whiteSpace: 'nowrap',
     border: 0
   };
   return (
--- a/src/index.js
+++ b/src/index.js
@@ -1127,7 +1127,7 @@ export default class Carousel extends React.Component {
   }
 
   render() {
-    const { currentSlide, slideCount, frameWidth } = this.state;
+    const { currentSlide, slideCount, frameWidth, hasInteraction } = this.state;
     const {
       disableAnimation,
       frameOverflow,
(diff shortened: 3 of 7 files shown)


XavierMeunier:
First, Thank you @sarmeyer for bringing more accessibility to Nuka Carousel  🙌

I'm curious to know why do you change `ul` & `li` to `div` ?

In the same PR, would it be possible to add `aria-selected={isActive}` L242 in `default-control.js`?

In case you don't know about it, there is a lot of recommandation about carousels accessibility [here](https://www.w3.org/TR/wai-aria-practices/#carousel).
If you're open to contributions in this field it would be a pleasure to contribute.

sarmeyer:
@XavierMeunier absolutely open to contributions! I'll make sure to add `aria-selected` to this PR

PR #804, opened 2021-10-18
fix: prevent initializeCarouselHeight loop to fix browser crash

### LINKED ISSUE #798: BIG timer problem
if you leave the page with the carousel for a while in the open tab and return, for example, after 3-4 hours to the page, it hangs tightly !! 

Reason found

photo_2021-09-19 16 16 22

**please fix it**

hsource:
#### Background

On a page that has an instance of nuka-carousel inside a hidden or 0-height div, the `setTimeout` in `initializeCarouselHeight` will infinitely loop since the measured slide height will always be 0.

When the user leaves the tab, setTimeouts may not be executed in the background, and when they return to the tab, the tab will crash: see #798 for others who've encountered this.

#### Fix

Set a limit on how many times the timeout will run before giving up. I set this to 10, since it already seems like plenty, and by the 10th iteration, it'll wait for over 10 seconds already.

Fixes #798

#### Type of Change

### How Has This Been Tested?

I installed a version of the package on a website that was having the issue.

I then ran this code in Chrome console to log any setTimeout calls.

```js
const origSetTimeout = window.setTimeout;
window.setTimeout = (...args) => { console.log('Called setTimeout', args); console.trace(); origSetTimeout(...args) }
```

I verified that the page stopped showing new calls to "Called setTimeout" after a minute or so

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -222,7 +222,8 @@ export default class Carousel extends React.Component {
   }
 
   initializeCarouselHeight() {
-    const heightCheckDelay = 200;
+    const initialDelay = 200;
+    let timesChecked = 0;
     const initializeHeight = (delay) => {
       this.timers.push(
         setTimeout(() => {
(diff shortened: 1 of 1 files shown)


sarmeyer:
Thanks for putting this fix in! I'll merge and release today or tomorrow

PR #820, opened 2021-12-08
[v5.0.0] Replace position absolute with width and translate

ValGeorgiev:
### Description
Remove absolute positioning of slides and use width and translate in the slider-list. Also add additional slides in front and back so we can swipe easily with animation when we use carousel with wrapAround. These slides that are just helpers currently are tripling the slides on the page, but this will be reduced in next PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,6 +85,7 @@
     "prettier-fix": "prettier \"**/*.{js,json,ts,tsx,css,md}\" --write",
     "preversion": "npm run check",
     "start": "webpack-dev-server --mode development",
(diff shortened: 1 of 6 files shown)


KenanYusuf (approved):
Nice! potentially something for a follow-up PR but we'll need to account for vertical carousels too https://61b07a202e5062004ae8fb5a-ptffapjxar.chromatic.com/?path=/story/nuka-carousel-carousel--vertical

ValGeorgiev:
Not needed anymore, closing

Issue #778, opened 2021-05-28
AnnounceSlide doesn't work if autoplay=true


JoReavell:
# Bugs and Questions

### Prerequisites

**Feel free to delete this section if you have checked off all of the following.**

- [x] I've searched open [issues](https://github.com/FormidableLabs/nuka-carousel/issues) to make sure I'm not opening a duplicate issue
- [x] I have read through the docs before asking a question
- [x] I am using the latest version of nuka-carousel

### Describe Your Environment

- What version of nuka-carousel are you using? 4.7.8
- What version of React are you using? 16.9.0
- What browser are you using? Chrome

### Describe the Problem

AnnounceSide message doesn't work when autoplay is true but autoplaying has been stopped by hovering over slide or using next/previous buttons.

**Expected behavior:** When autoplay functionality is stopped and set to false (e.g. by hovering over a slide or using next/previous buttons or paging-dots) I would expect the AnnounceSlide message to be available in the dom with `aria-live=polite`. I would expect the AnnounceSlide message to always be in the DOM with the `aria-live` attribute switching between `off` and`polite` as the autoplay attribute changes. 
**Actual behavior:** Currently when a screenreader user uses the next/previous or navigation buttons there is no indication that anything has changed as the AnnounceSlide method is never in the DOM if `autoplay=true` to begin with.

sarmeyer:
Hey @JoReavell thank you for submitting this issue! I spent some time on this yesterday. it isn't as easy of a fix as I was hoping. I am doing some refactoring that I hope will allow us to implement this properly.

vaishal:
Fixed in v5, will close this when it is released.

ValGeorgiev:
Hi @JoReavell, I believe this issue is fixed and released with v5. I will close it for now, feel free to reopen it you have any questions.

PR #832, opened 2022-01-18
AnnounceSlide - update aria-live value depending on autoplay

vaishal:
### Description

Always render `AnnounceSlide` and update the `aria-live` value depending on the `autoPlay` and `pause` values.

Fixes https://github.com/FormidableLabs/nuka-carousel/issues/778


ValGeorgiev (approved):
Looks great! 🌟

PR #841, opened 2022-01-26
Drag improvements

vaishal:
### Description

Focused on improving drag/swipe behaviour and added a drag threshold.

- Fixed slide jumping on initial drag. This bug prevented controls in a slide from being clicked.
- Added drag threshold. Once passed, the drag is ended and navigates to the next/prev slide. If not passed the slide animates back to its original position.
- Improved dragging first/last slides and fixed a bug with not being able to drag the last slide when `wrapAround` is false.

### DIFF
--- a/src-v5/new/carousel.tsx
+++ b/src-v5/new/carousel.tsx
@@ -56,6 +56,8 @@ const Carousel = (props: CarouselProps): React.ReactElement => {
     // boundary
     if (props.wrapAround || currentSlide < count - props.slidesToShow) {
       moveSlide(currentSlide + slidesToScroll);
+    } else {
+      moveSlide();
     }
   };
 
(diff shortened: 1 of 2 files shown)


src-v5/new/carousel.tsx
```diff
@@ -191,119 +206,72 @@ const Carousel = (props: CarouselProps): React.ReactElement => {
   };
 
   const onTouchEnd = () => {
-    if (!props.dragging) {
+    handleDragEnd();
```
ValGeorgiev, line 209: should we remove onTouchEnd function at all and use handleDragEnd function. One function less

ValGeorgiev (commented):
Looks great, I have one minor comment only. Also is it possible to add some tests that cover such functionality (not sure)

src-v5/new/carousel.tsx
vaishal: True could move it within an inline function.

ValGeorgiev (approved):

PR #857, opened 2022-03-01
Fix/carousel aria label

ValGeorgiev:
### Description

Add property to customise the frame aria label attribute of the carousel.


carlos-kelly (approved):
This is great. Thank you!

carlos-kelly:
Looks like we just need to run prettier then this is good to merge.

PR #860, opened 2022-03-06
Add support for event handlers for drag end and on drag

carlos-kelly:
### Description

Currently the drag start handler gets called for all drag movements. This PR separates out drag start, onDrag, and drag end into 3 different handlers. This lets the consumer add interactivity or disable interactivity when the user is swiping. 

#### Type of Change

### How Has This Been Tested?

Using the Next example and ensuring the correct methods are called the correct number of times.

- dragStart should only ever get called once per drag
- onDrag should get called several times as the user swipes
- dragEnd should only ever get called once per drag

Screen Shot 2022-03-06 at 3 23 49 PM

### DIFF
--- a/src-v5/carousel.tsx
+++ b/src-v5/carousel.tsx
@@ -212,10 +212,13 @@ export const Carousel = (props: CarouselProps): React.ReactElement => {
     };
   }, []);
 
-  const handleDragEnd = () => {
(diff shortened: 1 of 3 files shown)


ValGeorgiev (approved):
Nice one

KenanYusuf (approved):

PR #864, opened 2022-03-14
fix: ensure after slide index is correct

carlos-kelly:
### Description

Fixes `afterSlide` to give us the correct slide index that is normalized for the number of slides.

https://user-images.githubusercontent.com/1738349/158188643-29281779-cb09-45cf-966e-07d32b7b11ec.mov

#### Type of Change

### DIFF
--- a/src-v5/carousel.tsx
+++ b/src-v5/carousel.tsx
@@ -1,6 +1,6 @@
 /* eslint-disable complexity */
 
-import React, { useEffect, useState, useRef } from 'react';
+import React, { useEffect, useState, useRef, useCallback } from "react";
 import Slide from './slide';
 import { getSliderListStyles } from './slider-list';
 import { CarouselProps, KeyCodeFunction } from './types';
@@ -41,14 +41,8 @@ export const Carousel = (props: CarouselProps): React.ReactElement => {
 
   const carouselRef = props.innerRef || carouselEl;
 
-  const moveSlide = (to?: number) => {
-    const [slide] = getIndexes(
-      currentSlide,
-      currentSlide - slidesToScroll,
-      count
-    );
-
-    const nextIndex = () => {
+  const getNextIndex = useCallback(
+    (to?: number) => {
       const index = to ?? currentSlide;
       if (index < 0) {
         return index + count;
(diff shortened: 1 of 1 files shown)


src-v5/carousel.tsx
```diff
@@ -49,12 +49,20 @@ export const Carousel = (props: CarouselProps): React.ReactElement => {
...
     setCurrentSlide(to ?? currentSlide);
     setTimeout(
       () => {
         if (!isMounted.current) return;
-        to && props.afterSlide(currentSlide);
+
+        let nextIndex = to ?? currentSlide;
+        if (nextIndex < 0) {
+          nextIndex += count;
+        }
```
carlos-kelly, line 60: if it's less than `0` that means we went backwards so we want to add the total count to the negative number to get the right index.

src-v5/carousel.tsx
```diff
@@ -49,12 +49,20 @@ export const Carousel = (props: CarouselProps): React.ReactElement => {
...
+        if (nextIndex === count) {
+          nextIndex = 0;
+        }
```
carlos-kelly, line 63: if the next index is the total number of slides that means we went backwards then forwards (while looping) which means we are at the first slide.

src-v5/carousel.tsx
```diff
@@ -49,12 +49,20 @@ export const Carousel = (props: CarouselProps): React.ReactElement => {
...
+
+        typeof to === 'number' && props.afterSlide(nextIndex);
```
carlos-kelly, line 65: previously just doing the check for `to` would actually fail it if the slide index is `0` since that is normalized to `false`.

ValGeorgiev: I think the same thing will happen for `beforeSlide`. Could we fix this as well?

ValGeorgiev (approved):
Nice catch @carlos-kelly. Thanks for the PR as well. One minor comment from me. I tested it in several cases including when we have slidesToScroll set to 2 or 3 and it works fine. 

I am not big fan of declaring variables with `let`, maybe we can do something like this as well:
```
const getNextIndex = (index: number, count: number): number => {
  if (index < 0) {
    return index + count;
  }
  if (index === count) {
    return 0
  }
  return index
}

// in the moveSlide function
const nextIndex = getNextIndex(to ?? currentSlide, count)
```

Your approach is totally working and it's up to you, just proposing :D Thanks again for the PR

ValGeorgiev (approved):

PR #868, opened 2022-04-05
v5: Migrate scrollPage: reminded property to v5

ValGeorgiev:
### Description

As it was removed firstly, we are migrating `scrollPage: reminded` back to v5. Currently if we have v5 carousel without `wrapAround` and slides are not even to slidesToShow, we can see white screen in the end of the carousel. This can be prevented by adding scrollPage: reminded property.

This PR covers the straightforward scenario when slidesToShow and slidesToScroll are equal. There is no point slidesToScroll is more than slidesToShow, because some slides will be missed. Next PR will cover the scenario when slidesToScroll is less than slidesToShow.

### DIFF
--- a/src-v5/types.ts
+++ b/src-v5/types.ts
@@ -41,6 +41,11 @@ export interface Slide {
   offsetHeight: number;
 }
 
+export enum ScrollMode {
+  page = 'page',
+  reminder = 'reminder',
+}
+
 export enum D3EasingFunctions {
   EaseLinear = 'easeLinear',
   EaseQuad = 'easeQuad',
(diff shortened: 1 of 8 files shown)


src-v5/types.ts
```diff
@@ -41,6 +41,11 @@ export interface Slide {
   offsetHeight: number;
 }
 
+export enum ScrollMode {
+  page = 'page',
+  reminder = 'reminder'
```
embarks, line 46:
```suggestion
  remainder = 'remainder'
```

ValGeorgiev: hah silly me :D thanks @embarks

PR #890, opened 2022-04-22
Make explicit children prop for React 18 @types/react issue

### LINKED ISSUE #889: `@types/react` requires explicit children in props starting with v18
# Bugs and Questions

### Describe Your Environment

- What version of nuka-carousel are you using?
  5.0.13
- What version of React are you using?
  18.0.0
- What browser are you using?
  Irrelevant, but Microsoft Edge (Chromium)

### Describe the Problem
As of `@types/react@18.0.0` children is no longer included as an implicit prop, which causes issues for Carousel as `children` is not an explicit prop in `CarouselProps` in index.d.ts.

I created a code sandbox, but it doesn't replicate the issue I'm seeing because I don't think it uses TypeScript quite the same as a build process would. 

Instead using https://github.com/riceboyler/nuka-types and opening index.tsx (after running `yarn`) should demonstrate the Type errors.

**Expected behavior:** No errors from VSCode or my build process related to Carousel not having children as an available prop
**Actual behavior:** 

image

riceboyler:
### Description

Simply adds `children` as an explicit prop to `CarouselProps` in `index.d.ts`, since that's what TypeScript is using as its type definition.

Fixes #889 

#### Type of Change

#### Testing

Tested with installing this version on https://github.com/riceboyler/nuka-types and it resolved the Typescript error.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -147,6 +147,11 @@ export interface CarouselProps {
   cellSpacing?: number;
 
   /**
+   * Explicit children prop to resolve issue with @types/react v18
+   */
+  children?: React.ReactNode;
+
+  /**
    * Additional className
    */
   className?: string;


ValGeorgiev (approved):
awesome, thanks for this PR

PR #896, opened 2022-05-02
Refactor Carousel component, reconfigure linting and update deps

zubhav:
### Description

**Refactor Carousel components**
- Extracted all props at top of component to avoid repeated object access via `props.xyz`
- Wrapped component functions in `useCallback` with correct deps
- Fixed `useEffect` usage by supplying necessary deps
- Avoided memory leaks by implementing missing cleanup functions for timeouts

**Reconfigured linting**
- Updated the linting setup to be inline with changes to React ecosystem (previous setup was from 4 years ago)
- Introduced eslint-plugin-react-hooks to display errors and warnings for exhaustive-deps rule

**Upgraded deps**
- Eslint, eslint-config-prettier, eslint-plugin-react

Screenshot 2022-05-03 at 11 10 45

#### Type of Change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

### How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

### Checklist: (Feel free to delete this section upon completion)

- [x] My code follows the style guidelines of this project (I have run `yarn lint`)
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes (I have run `yarn test` and `yarn test-e2e`)
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] For any breaking API changes, I have updated the type definitions in [`index.d.ts`](../index.d.ts)
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published in downstream modules

PR #903, opened 2022-05-31
fix: Simplify and make `adaptiveHeights` work without suppressing eslint rule

hsource:
#### Motivation 

This is a small addition to #896 that fixes it so:

1. `adaptiveHeights` works
2. We don't need to suppress `react-hooks/exhaustive-deps`

This also simplifies the logic around `adaptiveHeights`. `adaptiveHeights` didn't work because the dependency in the `useEffect` was `visibleHeights`. `visibleHeights` is a ref, so it never changes, and never causes the `useEffect` to rerun.

This change makes it a state variable.

#### Changes

1. Instead of using a `ref` for `visibleHeights`, use `useStateWithRef` instead to create a state variable that matches a ref. When a `ref` changes, it doesn't cause the components to rerender, which caused the problem when using `adaptiveHeights`. This change ensures we always have the latest value 
2. Refactor the logic to update the `frameHeight` out of Slide into Carousel. It didn't make sense for every slide to update the frame height, when it really needed to be done once

#### Type of Change

### How Has This Been Tested?

1. Added new Storybook stories for `adaptiveHeight`
3. Test with storybook stories

Adaptive heights


ValGeorgiev:
Hey @hsource, thanks a lot for your PRs, I will need some time but definitely will go through them this week. Thanks again

src-v5/carousel.tsx
```diff
@@ -613,6 +620,63 @@ export const Carousel = (props: CarouselProps): React.ReactElement => {
   );
 };
 
+/**
+ * Adjust the frame height based on the visible slides' height if
+ * `adaptiveHeight` is enabled. Otherwise, just returns `auto`.
+ */
+const useFrameHeight = ({
```
ValGeorgiev, line 627: Could we move this hook in hooks.ts, please. The carousel.tsx is huge already

src-v5/slide.tsx
```diff
@@ -135,6 +129,7 @@ const Slide = ({
   const slideRef = useRef<HTMLDivElement>(null);
 
   // eslint-disable-next-line complexity
```
ValGeorgiev, line 131: Could we remove this line now, when the useEffect is simpler

ValGeorgiev (changes requested):
I had a quick look into the PR, it's great to have this eslint suppressing issue fixed for sure and thank you very much for your effort. I added 2 minor comments. Also I tested the carousel behaviour manually and found a bug. I think it's going to the default frameHeight value which is 0 and it's hiding the slide in some point (needs more investigation).

Here is the video. Carousel parameters:
```
"wrapAround":true
"adaptiveHeight":true
```

https://user-images.githubusercontent.com/12056467/171642394-e368fdf3-612a-4b40-8a85-9682c25e2eb7.mov

ValGeorgiev:
oops, this was automatically closed when I merged the other PR, I would love to have this PR merged in nuka!

PR #909, opened 2022-06-04
fix: Simplify and make `adaptiveHeights` work without suppressing eslint rule

hsource:
#### Motivation 

This is a small addition to #896 that fixes it so:

1. `adaptiveHeights` works
2. We don't need to suppress `react-hooks/exhaustive-deps`
3. `adaptiveHeights` doesn't just render a blank page with server-side-rendering

This also simplifies the logic around `adaptiveHeights`. `adaptiveHeights` didn't work because the dependency in the `useEffect` was `visibleHeights`. `visibleHeights` is a ref, so it never changes, and never causes the `useEffect` to rerun.

This change makes it a state variable.

#### Changes

1. Instead of using a `ref` for `visibleHeights`, use `useStateWithRef` instead to create a state variable that matches a ref. When a `ref` changes, it doesn't cause the components to rerender, which caused the problem when using `adaptiveHeights`. This change ensures we always have the latest value 
2. Refactor the logic to update the `frameHeight` out of Slide into Carousel. It didn't make sense for every slide to update the frame height, when it really needed to be done once
3. Change the `adaptiveHeights` so that it works with server-side-rendering, by introducing new `initializedAdaptiveHeight` variable
   - When `initializedAdaptiveHeight` is false: the frame has height `auto`; visible slides have height `auto`; invisible slides have height `0`
   - The client sets `initializedAdaptiveHeight` to true once we've measured all the visible slides' heights
   - When `initializedAdaptiveHeight` is true: the frame has height set to the tallest visible slide; all slides have height `100%`

#### Type of Change

### How Has This Been Tested?

1. Added new Storybook stories for `adaptiveHeight`
   - This includes a new Storybook story using `renderToString` for server-side-rendering testing
3. Test with storybook stories

Adaptive height


hsource:
@ValGeorgiev this is a recreated version of #903 and is ready for review/merge!

ValGeorgiev:
> @ValGeorgiev this is a recreated version of #903 and is ready for review/merge!

Hey @hsource, thanks for this, but I am still able to reproduce the mentioned issue [here](https://github.com/FormidableLabs/nuka-carousel/pull/903#pullrequestreview-993536907 )

hsource:
@ValGeorgiev Whoops! Sorry, I missed those comments. I pushed a new commit fixing all the issues. Here's a gif with the new Storybook story for adaptiveHeights with wrapAround:

customIndex

hsource:
@carlos-kelly @gksander - would it be possible to take a look at this PR? This PR ensures that `adaptiveHeights` works with server-side rendering (SSR) and also simplifies the logic.

packages/nuka/src/carousel.tsx
```diff
@@ -595,7 +600,7 @@ export const Carousel = (rawProps: CarouselProps): React.ReactElement => {
           width: '100%',
           position: 'relative',
           outline: 'none',
-          height: adaptiveHeight ? `${frameHeight}px` : 'auto',
+          height: frameHeight,
```
gksander, line 603: Nice. I wonder if we should also add a default transition for this `height` property so that the height change isn't so sudden. Something like:

```
transition: 'height 300ms ease-in-out',
willChange: 'height',
```

I see that this can be handled manually by passing in style props, but feels like it's worth having a sane default here.

gksander (approved):
This looks awesome – I appreciate the attention to SSR!

carlos-kelly (approved):
This is really great. I also like the new hook that saves state without forcing a render. Finally thank you for adding stories for this.

packages/nuka/src/carousel.tsx
carlos-kelly: I agree with making the transition smoother. This can be added in a follow up as well, but if we do add animations as default we should document it as well with how to disable it because it could be a surprise to users of the library.

gksander:
@hsource Looks like CI got hung up, we pushed up a small change in #935 that should run the full CI suite on PRs from forks. I'd love to get this merged, but we have two options:

1. You can rebase with/merge in `main` and push that up, and hopefully that'll run all of the CI suite and we can get this branch formally passing all CI checks, at which point you can squash+merge.
2. We can override the CI checks and force squash+merge for you. Happy to do this, but then I believe the commit is authored under my name and not yours, and I'd rather give credit where credit is due!

Let me know what you want to do, and we'll try to get your handful of (super helpful) PRs merged in ASAP.

hsource:
Thanks for the feedback! I'll take a look and make the fixes later today

packages/nuka/src/carousel.tsx
gksander: Good callout, Carlos! So let's leave this be for now and hit it as a follow-up task, since this PR already has a significant amount of work in it.

hsource:
I added the frame height animation, tested it with a new Storybook story, and added documentation to README.md

I also rebased on the latest main - hope that helps!

Animation

@gksander

PR #911, opened 2022-06-14
Render controls before slides in the carousel DOM

vala:
### Description

This change adds a new `renderControlsBeforeSlides` prop on the main `<Carousel/>` component, which when set to `true`, renders the controls before the slide frame in the DOM.

This allows screen reader and other keyboard users to first access the slider controls before the slider frame and its possible focusable contents, when tabbing through the web page the carousel is displayed on.

I didn't manage to find a WCAG 2.0 rule associated to the carousel control focus order, but we were asked to do so for a11y reasons by a partner company.

Existing W3C WAI-ARIA carousel examples ([here](https://www.w3.org/WAI/ARIA/apg/example-index/carousel/carousel-1-prev-next.html) and [also here](https://www.w3.org/WAI/ARIA/apg/example-index/carousel/carousel-2-tablist.html)), and other accessible carousel we found on the web seem to always place the controls before the carousel frame and contents.

As there doesn't seem to be an actual official rule about it, I just added the prop and made it default to `false`.

#### Type of Change

### How Has This Been Tested?

The non-e2e test suite doesn't seem to run for me, and when checking the code, it seems to call nuka-carousel v4 code (from the `src/` folder and not the `src-v5/` one).

So my tests where : 

```bash
$ yarn
$ yarn start:next
$ open http://localhost:3000/?params={"renderControlsBeforeSlides": true}
```

You can test it with the [PR Review APP](https://nuka-carousel-git-fork-respublica-conseil-243ed8-nuka-carousel.vercel.app/?params={%22renderControlsBeforeSlides%22:%20true})

Then tab through the page : the controls are focused first, the carousel frame after.

- [ ] New and existing unit tests pass locally with my changes (I have run `yarn test` and `yarn test-e2e`)

Note : I couldn't run the tests with `yarn run test`, so I didn't added tests, but updated the README. But the cypress tests seemed to all pass as green.
Plase let me know if there's any way to run the tests and if you'd like me to add some tests for the new prop.

Thanks !

### DIFF
--- a/src-v5/carousel.tsx
+++ b/src-v5/carousel.tsx
@@ -562,6 +562,16 @@ export const Carousel = (props: CarouselProps): React.ReactElement => {
         })}
       />
 
+      {renderControls(
+        props,
+        count,
+        currentSlide,
+        moveSlide,
+        nextSlide,
+        prevSlide,
+        slidesToScroll
+      )}
+
       <div
         className={['slider-frame', className || ''].join(' ').trim()}
         style={{
@@ -605,15 +615,6 @@ export const Carousel = (props: CarouselProps): React.ReactElement => {
           {wrapAround ? renderSlides('next-cloned') : null}
         </div>
       </div>
-      {renderControls(
-        props,
-        count,
-        currentSlide,
-        moveSlide,
-        nextSlide,
-        prevSlide,
-        slidesToScroll
-      )}
     </div>
   );
 };
--- a/src-v5/control-styles.ts
+++ b/src-v5/control-styles.ts
@@ -1,18 +1,23 @@
 import { CSSProperties } from 'react';
 import { Positions } from './types';
 
+const commonStyles: CSSProperties = {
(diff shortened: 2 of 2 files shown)


ValGeorgiev:
Hey @vala, Thank you for doing this, the fix is all good, but I don't think this should sit behind a property. Sounds more like the default behaviour that we want. We want to be accessibility first by default. Could you remove the property and make it by default and I will have another look and ship it to npm. Thanks again

vala:
Hi @ValGeorgiev thanks for your quick answer and sorry for the delay of my reply.

I updated the code to remove the parametric aspect of the feature and make it default behavior.

Thanks for your review !

ValGeorgiev (approved):

ValGeorgiev:
This should be in the nuka-carousel@5.1.2

vala:
Nice ! Thanks for the merging !

PR #933, opened 2022-07-20
Migrate to PNPM Workspace

### LINKED ISSUE #930: [v6] Converge examples, move to (pnpm) workspace to share core with examples
There are currently two example apps in Nuka:

- A webpack-based sample that lives in `/demo`, and appears to be a bit dated.
- A Next.js app that lives in `examples/nextjs`. This app has its own set of dependencies from the rest of the codebase, but "reaches out" to grab nuka source files for its own use.

We should tackle the following work to clean this up a bit:

- [ ] We should ditch the webpack-based sample (in `/demo`) and move that demo sample into the `examples/nextjs` app.
- [ ] We should use pnpm workspaces so that the example shares dependencies with the rest of the repo's code, and can use compiled nuka source code (which is useful to ensure the build process works as expected in a real app without having to make a release).

carlos-kelly:
### Description

Migrate the repository to use PNPM workspaces for the NextJS demo app and the core library.

Fixes #930 

#### Type of Change


packages/nuka/src/utils.ts
```diff
@@ -33,10 +33,6 @@ export const addEvent = (
   }
   if (elem.addEventListener) {
     elem.addEventListener(type, eventHandler, false);
-  } else if (elem.attachEvent) {
-    elem.attachEvent(`on${type}`, eventHandler);
-  } else {
-    elem[`on${type}`] = eventHandler;
```
carlos-kelly, line 39: These are Internet Explorer methods that don't exist in modern APIs. We could polyfill the types but I think it's better to drop.

examples/nextjs/package.json
```diff
@@ -10,10 +10,14 @@
   },
   "dependencies": {
     "next": "^12.1.4",
-    "wicg-inert": "^3.1.1"
+    "wicg-inert": "^3.1.1",
+    "nuka-carousel": "workspace:nuka-carousel@*"
```
ryan-roemer, line 14: I think this should just be `"*"`? PNPM in workspace mode should auto-infer this already and we otherwise have to check this works correctly with `changesets version`

CONTRIBUTING.md
```diff
@@ -32,17 +35,17 @@ Basics:
...
-$ yarn run test
+$ pnpm run lint
+$ pnpm run test
 ```
 
 And E2E tests (you _must_ be on `node@8+`):
 
 ```sh
-$ yarn run test-e2e
+$ pnpm run test-e2e
```
ryan-roemer, line 48: I think this should be:

```suggestion
$ pnpm run test
```

which then runs the in package `test:e2e`

examples/nextjs/package.json
carlos-kelly: I changed it to this then realized it was using the pnpm cache. I think I need to figure out how to link it to `packages\nuka\src\index.tsx` over `lib` or any other output so developers can iterate on Nuka and have the source changes show up in the working demo.

ryan-roemer: Yep. So, reading up, I think this still _should_ work as we have it. But. here's an `.npmrc` option that forces the link in my check out of this repo with just `"*"` as the specifier:

```diff
diff --git a/.npmrc b/.npmrc
index 319e41e..a018756 100644
--- a/.npmrc
+++ b/.npmrc
@@ -1 +1,2 @@
 strict-peer-dependencies=false
+prefer-workspace-packages=true
```

gksander (commented):
This looks awesome. Dev workflow feels really good. Only thing that I found that needs a bit of tweaking is the release/npm flow. (Dropped some deets in Slack.)

gksander (approved):
Nice, looks good to me!

ryan-roemer (approved):
Looks great!!!

PR #941, opened 2022-08-02
fix: Correct number and indices of dots in remainder scrollMode

fritz-c:
### Description

Previously, when `slidesToScroll` was set to a value larger than `1` with `scrollMode` set to `"remainder"`, the dots shown were missing the last one. In the case of fractional `slidesToShow` values, the index of the last dot was also incorrect.

| image |
|---|
| _Example: 2 dots, not the expected 3, shown for 5 slides with `slidesToScroll={2}` [Demo link](https://nuka-carousel-next.vercel.app/?slides=5&params=%7B%22scrollMode%22:%22remainder%22,%22slidesToShow%22:2,%22slidesToScroll%22:2%7D)_ |

This PR fixes both these issues and adds a few unit tests to double-check the function that generates the dot indices.

#### Type of Change

### How Has This Been Tested?

I created unit tests that passed a variety of parameters to the `getDotIndexes` function to test that it returned indices as expected.

The issue can be reproduced by looking at the dots versus the button behavior with this [issue example](https://nuka-carousel-next.vercel.app/?slides=5&params=%7B%22scrollMode%22:%22remainder%22,%22slidesToShow%22:2,%22slidesToScroll%22:2%7D).


packages/nuka/src/default-controls.tsx
```diff
@@ -190,22 +185,12 @@ export const PagingDots = (props: ControlProps) => {
     []
   );
 
-  const indexes = useMemo(
```
fritz-c, line 193: The identity of the array here wasn't important, nor is `getDotIndexes` particularly costly, so I took out the memoization.

gksander (approved):
This is awesome. And thank you for getting some testing in place!

PR #945, opened 2022-08-03
Add changeset for previous bugfix PRs

fritz-c:
### Description

Adds a changeset corresponding to the last two bugfix PRs I merged, which did not have changesets associated with them at the time.


gksander (approved):
Dope. Let's merge this and see if our GH actions are working as expected 😅

PR #947, opened 2022-08-03
fix: add slide-visible classes appropriately when slideIndex has fractional digits

fritz-c:
### Description

When the slideIndex has fractional digits (e.g., 0.5, 1.2), `slide-visible` classes are not being applied correctly.

[Bug demo](https://nuka-carousel-next.vercel.app/?slides=5&params=%7B%22scrollMode%22:%22remainder%22,%22slidesToShow%22:2.5,%22slidesToScroll%22:1%7D): click on `next` until you reach the end of the carousel and check slide 3's classes. It is missing `slide-visible` even though it is visible.

This PR fixes that.

#### Type of Change

### How Has This Been Tested?

I added unit tests for the `isSlideVisible` function.

### DIFF
--- a/.changeset/pink-apricots-pay.md
+++ b/.changeset/pink-apricots-pay.md
@@ -1,0 +1,5 @@
+---
+'nuka-carousel': patch
+---
+
+fix missing slide-visible classes when slideIndex has fractional digits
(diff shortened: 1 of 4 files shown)


gksander (approved):
Awesome, keep up the great work!

PR #948, opened 2022-08-03
Update contribution-related docs

fritz-c:
### Description

This makes the GitHub issue and PR templates a bit easier to use, putting in comments text that is only meant for the author and causes visual clutter when not proactively deleted. Most of the changes here are up for discussion, though there are a few changes of the `yarn lint` -> `pnpm run lint` variety that are not so opinionated.


gksander (approved):
🙇

PR #949, opened 2022-08-03
Use inline image src for demo

fritz-c:
### Description

The Chromatic UI tests in the CI, performing visual snapshot tests for the component, often snag on erroneous diffs caused by network issues when loading the sample slides from the demo site ([example Chromatic run](https://www.chromatic.com/test?appId=61b07a202e5062004ae8fb5a&id=62e94486780e4059ce4b4d5b)). This PR replaces the hosted image files with inline base64 representations of the same images, to be bundled in the site source code, hopefully avoiding the network flakiness issues.

#### Type of Change

CI-related

### How Has This Been Tested?

I ran the cypress tests, checked the demo site locally, and everything checked out. The hope is that CI visual checks on future PRs will be more stable with this change.


gksander (approved):
DOPE

PR #950, opened 2022-08-03
Make visual snapshot diffs more consistent

fritz-c:
### Description

The goal is the same as #949. Only, #949 attempted to fix it by changing the demo site, when visual regression tests are actually run against storybook. So this PR reverts #949 and attempts to fix the UI diff issue with a bit more traditional fix - getting rid of the storybook's dependency on a [3rd-party image placeholder service](https://via.placeholder.com/800/7732bb/ffffff/&text=slide1).


gksander (approved):
Awesome, thanks for getting to the bottom of this!

PR #952, opened 2022-08-05
fix: handle indices properly when `cellAlign` is `center` or `right`

fritz-c:
### Description

There are a variety of intertwined bugs related to slide index bounds (what the dots correspond to, also what determines whether prev/next buttons should be disabled) caused when `cellAlign` and `scrollMode` are thrown into the mix.

1. With `autoplayReverse` set to `true` and `cellAlign` set to something other than `left`, the component starts with an odd offset: [issue repro](https://nuka-carousel-next.vercel.app/?slides=5&params=%7B%22cellAlign%22:%22center%22,%22autoplayReverse%22:true,%22slidesToShow%22:2,%22slidesToScroll%22:2%7D). Note how the rightmost slide, which we should presumably start from, is partially concealed, and we cannot click the button to move over to it.

2. Autoplay does not go to the last slide with `cellAlign` of `center`: [issue repro](https://nuka-carousel-next.vercel.app/?slides=5&params=%7B%22cellAlign%22:%22center%22,%22autoplay%22:true,%22autoplayInterval%22:1000,%22slidesToShow%22:2,%22slidesToScroll%22:1%7D)

3. Scrolling to the right, we cannot reach slide 4: [issue repro](https://nuka-carousel-next.vercel.app/?slides=4&params=%7B%22cellAlign%22:%22right%22,%22slidesToShow%22:2,%22slidesToScroll%22:2%7D)

This PR fixes the above, as well as performs a bit of generic refactoring.

~~**NOTE**: the fixes change some `scrollMode="remainder"`-like behaviors that were baked-in before. For example, [this demo](https://nuka-carousel-next.vercel.app/?slides=6&params=%7B%22slidesToShow%22:2,%22slidesToScroll%22:2%7D) disables the "NEXT" button once the sixth slide is in view (e.g., `slideIndex = 4`).
With the fixes in this PR, the same scenario ([demo from PR preview](https://nuka-carousel-git-fix-handle-index-with-ce-f0dafc-nuka-carousel.vercel.app/?slides=6&params=%7B%22slidesToShow%22:2,%22slidesToScroll%22:2%7D)) would see the "NEXT" button enabled at that point, so the user could click again, resulting in slide 6 on the left side and whitespace on the right.
This _could_ be seen as a breaking change. I tend to see it as making the component more closely follow the behavior that the component props defines. Setting `scrollMode="remainder"` will produce the same result that was baked-in before, if that is what is desired.
This change reduces the assumptions we are making about the library user's use case. For example, without the fix, the `current-slide` class will never be assigned to the final slide in the scenario above, whereas it can be assigned with the fix in place.~~ This change was omitted, as the flexibility with `current-slide` I pitched here didn't really make sense when you were jumping two slides at a time (=skipping every other slide) anyway. Also, it seemed like a bit of a trap that you could have three slides shown per page with a collection of 9 slides, and still end up with 4 pages.

#### Type of Change

### How Has This Been Tested?

A variety of unit tests appearing in the PR, and manually


packages/nuka/cypress/integration/carousel.spec.js
```diff
@@ -89,12 +96,21 @@ context('Carousel', () => {
         .first()
         .should('have.attr', 'data-slide', 'Slide 5');
 
+      cy.get('button').contains('Prev').should('not.be.disabled');
+      cy.get('button').contains('Next').should('not.be.disabled').click();
+
+      cy.get('.slide.slide-visible')
+        .should('have.length', 1)
+        .find('img')
+        .should('have.attr', 'data-slide', 'Slide 6');
```
fritz-c, line 105: This modification corresponds to the change in behavior in the scenario laid out in the PR description.

packages/nuka/cypress/integration/infinity-carousel.spec.js
```diff
@@ -22,16 +26,13 @@ context('Infinity Carousel', () => {
       cy.get('button').contains('Prev').should('not.be.disabled');
       cy.get('button').contains('Next').should('not.be.disabled').click();
 
-      cy.wait(500);
```
fritz-c, line 25: Cypress will wait for animations to finish, so the `cy.wait`s are unnecessary:
https://docs.cypress.io/guides/core-concepts/interacting-with-elements#Animations

fritz-c:
Regarding the changes in starting index for right-aligned stories:
When working with `scrollMode="page"`+`cellAlign="left"`, this library prevents scrolling once all slides are reachable. So a carousel with 9 slides and `slidesToShow={3}` will only have three navigation dots, one for each "page", corresponding to indices `[0, 3, 6]`.
With the identical case, save for being flipped around via `cellAlign="right"`, we need to start from index 2 to similarly minimize unnecessary pages, for a result of `[2, 5, 8]`. Essentially, the algorithm to generate such an array works from the tail end of the slides, at `slideCount - 1`, moving towards the front until all slides have been reached.

This has the effect of creating some potentially puzzling offsets if you start from the leftmost end of the carousel, as is default.
[e.g., whitespace gap between the start of the slideshow and the leftmost side](https://61b07a202e5062004ae8fb5a-ovdptvdbni.chromatic.com/?path=/story/nuka-carousel-carousel--cell-align-right)

But when using the carousel from right-to-left, whether it be with `autoplayReverse` on or by setting `slideIndex` to the last slide initially, this initial index makes perfect sense.

gksander (approved):
Awesome!!!

PR #956, opened 2022-08-16
fix: more natural swiping behavior

fritz-c:
### Description

Swiping the carousel to the left and right using the mouse or touch gestures felt jittery and uncomfortable compared with native UI or horizontal scroll areas. This PR makes those gestures feel more like what users expect from native UI:

* using inertia calculations to enable quick flicks of the finger to scroll the slides that a simple `endPosition - startPosition > threshold` formula would normally ignore. Swiping multiple slides at a time is also now possible.
* The carousel now "sticks" to your finger regardless of whether you already exceeded the threshold, or touched in the middle of a sliding animation.

A couple bug fixes from this PR:

* The carousel width used in threshold detection was often stale, resulting in very little dragging needed to trigger carousel movement. Now the threshold works as expected.
* disableEdgeSwiping now works with cell alignments of `center` and `right`.
* setting `dragging={false}` will no longer disable carousel swiping on mobile (`swiping={false}` is the way to do that).

#### Type of Change

### How Has This Been Tested?

Running storybook locally and opening it on my phone, testing it in desktop browsers, new cypress tests.


packages/nuka/cypress/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
```
fritz-c, line 1: Verbatim from https://docs.cypress.io/guides/tooling/typescript-support#Configure-tsconfig-json

packages/nuka/src/slider-list.ts
```diff
@@ -93,45 +95,85 @@ const getPositioning = (
...
 
   return {
     width,
     textAlign: 'left',
+    userSelect: 'auto',
     transition:
-      isAnimating && slideAnimation !== 'fade'
-        ? `${speed || 500}ms ease 0s`
+      draggedOffset === 0 && isAnimating && slideAnimation !== 'fade'
+        ? `${speed}ms cubic-bezier(0.16, 1, 0.3, 1) 0s`
```
fritz-c, line 176: This is one of the iffier changes in the PR - changing the easing function used will make the slide changes look snappier / less relaxed than before. The goal of changing this was to make transitions starting from the end of a swipe look less stilted - `ease` has a slow buildup to its max speed which, coming off of a swipe, is unnatural in that it slows down and speeds up again. This new easing function does not have the slow buildup at the front, avoiding that issue, but, again, makes it feel "snappy" where it may not have before.

buchansm (approved):
The refactor work looks good 👍, and this definitely feels a lot snappier! Nice work.

gksander (approved):
🎈 LGTM

PR #958, opened 2022-08-17
Fix keyboard interactions

fritz-c:
### Description

The carousel did not animate when the keys corresponding to "firstSlide" or "lastSlide" were pressed. It also stopped at the wrong slide for certain `cellAlign` values. Click on the carousel and press the "Q" and "E" keys on [this repro demo](https://nuka-carousel-next.vercel.app/?slides=8&params=%7B%22slidesToScroll%22:3,%22enableKeyboardControls%22:true,%22cellAlign%22:%22center%22,%22slidesToShow%22:3%7D) to try for yourself. One more thing fixed was that key presses that were registered with the handler did not cause `preventDefault`/`stopPropagation` to be called, so both the carousel and the browser would respond to the same key event. To experience that, open the demo linked above, make the browser window short enough that you get a vertical scroll bar, click on the carousel, and then press the up and down arrow keys.

This PR fixes the behavior described above and simplifies the logic of the keyboard handler.

#### Type of Change

### How Has This Been Tested?

I added a simple integration test for the basic keyboard functionality. I also confirmed the behavior manually in storybook.

### DIFF
--- a/.changeset/four-olives-hope.md
+++ b/.changeset/four-olives-hope.md
@@ -1,0 +1,5 @@
+---
+'nuka-carousel': patch
+---
+
+keyboard-triggered "firstSlide" or "lastSlide" actions are now animated, and now take cellAlign into account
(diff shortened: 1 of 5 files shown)


buchansm (approved):
Nice, I like this a lot. ✅

PR #959, opened 2022-08-19
fix: carousel slide easing

### LINKED ISSUE #921: [v6] Carousel slide easing
Add support for configurable slide easing.

buchansm:
### Description

The user can now utilize the `easing` property for `<Carousel />`. The slider transitions via `d3-ease` interpolation methods.

Fixes #921

https://user-images.githubusercontent.com/20046764/185654429-93af39b7-806f-4cc0-a57e-406d928fd4a9.mp4

#### Type of Change

### How Has This Been Tested?

Running storybook locally and  testing it in desktop browsers.


packages/nuka/src/hooks/use-transition.ts
```diff
@@ -0,0 +1,48 @@
+import { useEffect, useState, useRef } from 'react';
+import { D3EasingFunctions } from 'src/types';
+import * as d3Ease from 'victory-vendor/d3-ease';
```
ryan-roemer, line 3: We have victory vendor as a development dependency, but this is production usage. It’s also worth having a discussion of whether for production victory vendor is appropriate because it brings in a lot of extra unrelated D3 packages. For frontend builds, tree shaking should take care of most of the extra stuff, but it does add extra stuff to like a node server side package

package.json
```diff
@@ -47,6 +47,7 @@
     "react-dom": "^18.0.0",
     "shx": "^0.3.4",
     "start-server-and-test": "^1.14.0",
-    "typescript": "^4.5.2"
+    "typescript": "^4.5.2",
+    "victory-vendor": "^36.6.3"
```
ryan-roemer, line 51: Is this package used at this level? We also have it in the Nuka package

buchansm: No it's not, I'll remove that. Nice catch!

packages/nuka/src/hooks/use-transition.ts
buchansm: Ah, I totally misunderstood. 

I was actually wondering, @gksander, what your thoughts are on taking my `useTransition()` hook and turning it into a lightweight animation library? And then I can create a `d3-ease` vendor package inside of that library and we can remove `victory-vendor` entirely from `nuka-carousel`.

packages/nuka/src/hooks/use-transition.ts
```diff
@@ -0,0 +1,48 @@
...
+  const updateFrameLoop = () => {
+    frameLoop.current = setInterval(() => {
+      const currentTime = Date.now();
+      const normalizedTime = Math.min(
+        1,
+        (currentTime - startTime.current) / duration
+      );
+      setNormalizedTime(normalizedTime);
+      if (normalizedTime === 1) clearFrameLoop();
+    }, fps);
```
gksander, line 37: This type of thing normally concerns me a bit, since we're setting React state _a lot_ here. However, I think we're doing a similar thing with out dragging – so this probably isn't out of norm for our existing setup.

For a side note, in terms of React performance, it's usually best to:

- use a `requestAnimationFrame` loop instead of `setInterval` and perform similar logic.
- use a DOM reference and imperatively update the DOM property, as opposed to using React state – so that we don't have to go through 60 React DOM reconciliations per second.
- Once an animation finishes, _then_ you set React state to get the rest of your React logic up to date with your imperative DOM manipulations.

However, that's generally much more complex and less readable than using React state, so maybe not something we worry about here.

packages/nuka/src/slider-list.tsx
```diff
@@ -0,0 +1,231 @@
...
+          // the most negative translate value
+          `clamp(${clampOffsets[1]}%, ${draggedOffset}px, ${clampOffsets[0]}%)`
+        : `${draggedOffset}px`;
+
+      return `translate3d(${
+        draggedOffset ? clampedDraggedOffset : `${slideBasedOffset}%`
+      }, 0, 0)`;
+
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [transition === 1 ? draggedOffset : transition]);
```
gksander, line 204: Is it dangerous to have a dynamic dependency list here?

gksander (commented):
Still working my way through this, but wanted to drop a couple initial comments.

packages/nuka/src/slider-list.tsx
buchansm: It was essential for preventing duplicate positioning and still allow the initial drag event to fire.

packages/nuka/src/hooks/use-transition.ts
buchansm: I could try to implement it this way in a follow up PR.

fritz-c: RE: Grant's feedback: I did a quick check of [how popular animation libraries set up their APIs](https://blog.logrocket.com/best-react-animation-library-top-7-libraries-compared/) to update the DOM, and it appears using React state / values in JSX to update is pretty standard, so I would assume it's decently performant as long as there's not too much logic/UI stuffed in there.

I have heard the `requestAnimationFrame` advice before, too, and usually rely on that method for anything related to animation.

fritz-c (changes requested):

packages/nuka/cypress/integration/swipe-carousel.spec.ts
```diff
@@ -110,7 +110,12 @@ context('Swipe Carousel', () => {
       cy.swipeSlider(carouselWidth);
 
       cy.get('.slider-list').should(($sliderList) => {
-        expect(getSliderXOffset($sliderList)).to.equal(-carouselWidth);
+        expect(getSliderXOffset($sliderList)).to.be.greaterThan(
+          -(carouselWidth + 1)
+        );
+        expect(getSliderXOffset($sliderList)).to.be.lessThan(
+          -(carouselWidth - 1)
+        );
```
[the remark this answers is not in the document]
buchansm: The variation is being caused by fractions of pixels I think in the [logic](https://github.com/FormidableLabs/nuka-carousel/pull/959/files#diff-71bdb43ae92624a80009148378a295d5bd5a5bbdd0a873f33d46bd891c2024f9R97). It ends up making it so the `sliderXOffset` is off by a fraction of a pixel. It's not visually noticeable but the numbers are just slightly off.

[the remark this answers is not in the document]
fritz-c: Any way to change the logic such that when we've reached the end state it bypasses that calculation and just returns the exact value it should be?

packages/nuka/src/slider-list.tsx
```diff
@@ -0,0 +1,231 @@
...
+    const [initialRect, setInitialRect] = useState<DOMRect>();
+    const rect = ref.current?.getBoundingClientRect();
+
+    const currentPosition = useMemo(() => {
+      if (rect && initialRect) {
+        return ((rect.x - initialRect.x) / rect.width) * 100;
+      }
+      return 0;
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [currentSlide, draggedOffset, initialRect]);
```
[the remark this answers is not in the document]
buchansm: Totally.

fritz-c (approved):

packages/nuka/src/hooks/use-transition.ts
fritz-c: I updated this hook (it had since been renamed to `useTween`) to use the standard `d3-ease` package instead of the `victory-vendor` one. I ended up reducing the flexibility of the hook by changing the generic `deps` array to a couple specific variables to have the internal `useEffect` watch. It was a trade-off I made for the sake of avoiding future bugs related to stale values, etc., and unfortunately makes it less of a candidate to become its own library.

fritz-c: Moving to `d3-ease` broke things due to [the same ESM issues that Victory had](https://formidable.com/blog/2022/victory-esm/). I will follow up on slack (#nuka-carousel) to discuss this.

buchansm:
These changes are brilliant @fritz-c! Thanks for taking the time to simplify/refactor this.

fritz-c:
My pleasure!

PR #971, opened 2022-09-14
Alternate easing api and `edgeEasing` prop

fritz-c:
### Description

Instead of bundling in the d3-ease functions, the easing prop now accepts a callback that accepts functions of a form identical to [d3-ease’s ease functions](https://github.com/d3/d3-ease). That way those who want to customize their easing function can, those who want to use d3-ease’s functions can (and benefit from tree-shaking of the ones they don't use), and those who are happy with the default don't have to bear the extra weight of added d3 ease functions. And we get to sidestep all the d3 ESM mess. The README now includes a code snippet showing how users might use the new API.

As I was working on that, I noticed there was an old easing prop, `edgeEasing` that hadn't been addressed yet, so I added that in, too.

And as I was working on that, I noticed that there's a bit of an animation bug in that edge swipes result in the carousel flashing into its final position for one frame before starting the animation, so I added a kind of rough workaround for that.

#### Type of Change

- [?] Breaking change (fix or feature that would cause existing functionality to not work as expected) **It would be a breaking change, save for the fact that we haven't released a version with `easing` available yet**

### How Has This Been Tested?

Manually

### DIFF
--- a/packages/nuka/src/default-carousel-props.tsx
+++ b/packages/nuka/src/default-carousel-props.tsx
@@ -3,11 +3,13 @@ import {
   Alignment,
   InternalCarouselProps,
   ControlProps,
-  D3EasingFunctions,
   ScrollMode,
 } from './types';
 import { NextButton, PagingDots, PreviousButton } from './default-controls';
 import { defaultRenderAnnounceSlideMessage } from './announce-slide';
+
+const easeInOut = (t: number) =>
+  t < 0.5 ? 4 * t * t * t : (t - 1) * (2 * t - 2) * (2 * t - 2) + 1;
 
 const defaultProps: InternalCarouselProps = {
   adaptiveHeight: false,
@@ -28,8 +30,8 @@ const defaultProps: InternalCarouselProps = {
   disableEdgeSwiping: false,
   dragging: true,
   dragThreshold: 0.5,
-  easing: D3EasingFunctions.EaseCircleOut,
-  edgeEasing: D3EasingFunctions.EaseElasticOut,
+  easing: easeInOut,
+  edgeEasing: easeInOut,
   enableKeyboardControls: false,
   frameAriaLabel: 'carousel-slider',
   keyCodeConfig: {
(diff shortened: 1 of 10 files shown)


gksander:
I really like this approach. I think it's great to not burden all users with d3 bloat – and just allow users who _do_ want to use d3 eases to provide it on their own.

packages/nuka/src/slider-list.tsx
```diff
@@ -125,6 +124,23 @@ export const SliderList = React.forwardRef<HTMLDivElement, SliderListProps>(
...
+        (currentSlide === dotIndexes[dotIndexes.length - 1] &&
+          animationDistance > 0));
+    const { value: transition, isAnimating } = useTween(
+      speed,
+      !isEdgeEasing ? easing : edgeEasing,
+      // animationDistance is assumed to be unique enough that it can be used to
+      // detect when a new animation should start. This is used in addition to
+      // currentSlide because some animations, such as those with edgeEasing, do
+      // not occur due to a change in value of currentSlide
+      currentSlide + animationDistance,
```
fritz-c, line 140: If I were to be a bit less hacky/ more verbose with the extra dependency on animationDistance, it would have doubled the code in some areas (see the ref storing and comparison in `useTween`), and the logic for each of the two dependencies would have been identical.

packages/nuka/src/default-carousel-props.tsx
```diff
@@ -3,12 +3,13 @@ import {
   Alignment,
   InternalCarouselProps,
   ControlProps,
-  D3EasingFunctions,
   ScrollMode,
 } from './types';
 import { NextButton, PagingDots, PreviousButton } from './default-controls';
 import { defaultRenderAnnounceSlideMessage } from './announce-slide';
 
+const easeOut = (t: number) => --t * t * t + 1;
```
gksander, line 11: Is the double negative intentional here? Also, is it fair game to do `t ** 3 + 1`?

gksander (approved):
Looks good! Couple minor questions, but I like this.

packages/nuka/src/default-carousel-props.tsx
fritz-c: The double negative is intentional. I found this in a gist, and I believe it's some code-golfing level syntax to do `(t - 1) * (t - 1) * (t - 1) + 1`. But I like your idea of using the exponentiation operator, so I'll switch it to that.

PR #972, opened 2022-09-22
Set draggable=false on only children of the SliderList, not the whole document

majgaard:
### Description

Nuka Carousel currently applies its `.slider-list img` query selector to the entire document, rather than the narrower scope it ought to impact.
In case of conflicting class names, Nuka Carousel may impact elements beyond its own scope.

#### Another less common reason:
I intend to use Nuka Carousel inside a React portal which portals its content into an `<iframe />` element. Because my React code runs on the host window, libraries which access the global `window` & `document` end up impacting *only the content outside the iframe*. The motivation behind portaling into an iframe is to shelter my UI from undue influence from the host page it is embedded on (and shelter the host page from things my code may be doing). If I depend on libraries which access the global `window` or `document`, things will break ¯\_(ツ)_/¯

#### Type of Change

- [ ] New feature (non-breaking change which adds functionality)
- [ ] This change requires a documentation update

### How Has This Been Tested?

I ran the whole test suite, and also verified that the `draggable=false` attribute was being set on the demo carousel images.

### Checklist

### DIFF
--- a/.changeset/metal-teachers-behave.md
+++ b/.changeset/metal-teachers-behave.md
@@ -1,0 +1,5 @@
+---
+'nuka-carousel': patch
+---
+
+Applies draggable=false only to children of sliderList, not entire document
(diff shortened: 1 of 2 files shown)


fritz-c:
This fix is now available in v5.4.0. Thanks!

Issue #969, opened 2022-09-12
[v6] Expand ref type, use React.forwardRef


fritz-c:
The prop `innerRef` does not allow the ref to point to a `null` value, even though that is the unavoidable initial value prior to first render. This will require library users to jump through some TypeScript hoops to make sure they can pass their ref to `innerRef`.
https://github.com/FormidableLabs/nuka-carousel/blob/df842370ff3cc1b4f95d6171a2998c31a608da0d/packages/nuka/src/types.ts#L330

One option is to convert the `innerRef` prop to a native-to-React `ref` prop by wrapping the carousel function in `React.forwardRef`, which, given the right element type, will factor in the nullability of the ref.

We could alternatively retain the `innerRef` and just expand the type, so we would be free to set a collection of imperative APIs ( e.g., `MyCarousel.nextSlide()`, `MyCarousel.goToSlide()`, etc) to the native `ref` via [useImperativeHandle](https://reactjs.org/docs/hooks-reference.html#useimperativehandle), though I feel like our existing API gives enough flexibility that the extra API surface would only cause confusion.

This is a v6 issue because renaming the prop would break people's code, and expanding the type could, too. For example, TypeScript code that has no null checks before calling something like `myCarouselInnerRef.current.getBoundingClientRect()` would suddenly start to get a type error about `getBoundingClientRect` not existing on `null`.

PR #981, opened 2022-10-26
Fix: use forwardref for nullability

JeehyaeDance:
### Description

This PR is for "[v6] Expand ref type, use React.forwardRef #969". Wrapped `carousel` component with `React.forwardRef` and replaced `innerRef` prop with a native-to-React `ref` prop.

Fixes # 969 (issue)

#### Type of Change
- [ ] This change requires a documentation update

### Checklist

- [ ] I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/packages/nuka/src/carousel.tsx
+++ b/packages/nuka/src/carousel.tsx
@@ -18,672 +18,682 @@ import {
 } from './utils';
 import { useFrameHeight } from './hooks/use-frame-height';
 import { getDotIndexes } from './default-controls';
+import { useForwardRef } from './hooks/use-forward-ref';
 
-export const Carousel = (rawProps: CarouselProps): React.ReactElement => {
-  /**
-   * We need this cast because we want the component's properties to seem
-   * optional to external users, but always-present for the internal
-   * implementation.
-   *
-   * This cast is safe due to the `Carousel.defaultProps = defaultProps;`
-   * statement below. That guarantees all the properties are present, since
-   * `defaultProps` has type `InternalCarouselProps`.
-   */
-  const props = rawProps as InternalCarouselProps;
+// eslint-disable-next-line react/display-name
+export const Carousel = React.forwardRef<HTMLDivElement, CarouselProps>(
+  (rawProps, ref): React.ReactElement => {
+    /**
+     * We need this cast because we want the component's properties to seem
+     * optional to external users, but always-present for the internal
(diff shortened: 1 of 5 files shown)


packages/nuka/src/hooks/use-forward-ref.ts
```diff
@@ -0,0 +1,17 @@
+import { useEffect, useRef, ForwardedRef } from 'react';
+
+export const useForwardRef = <T>(ref: ForwardedRef<T>) => {
```
JeehyaeDance, line 3: `const carouselRef = ref || defaultCarouselRef;`
At first, I tried to pass `ref` that was created from `React.forwardRef` to line 123 in `carousel.tsx`, however, there was a typescript error. (Property 'current' does not exist on type '(instance: HTMLDivElement | null) => void'.ts)
That's why I created this hook to handle null and I referenced it from [here](https://github.com/facebook/react/issues/24722).

packages/nuka/src/carousel.tsx
```diff
@@ -18,672 +18,682 @@ import {
...
-  const slideCount = filteredSlides.length;
-
-  const slidesToScroll =
-    animation === 'fade' ? slidesToShow : propsSlidesToScroll;
-
-  const [currentSlide, setCurrentSlide] = useState<number>(() =>
-    getDefaultSlideIndex(
+import { useForwardRef } from './hooks/use-forward-ref';
+
+// eslint-disable-next-line react/display-name
```
fritz-c, line 23: You can avoid having to disable this rule by naming the function argument like `function Carousel(rawProps, ref): React.ReactElement {`

packages/nuka/src/hooks/use-forward-ref.ts
fritz-c: Makes sense. Nice find 👍

fritz-c (changes requested):
You'll need to add a changeset. We might want to look into any other breaking changes for v6 to release at the same time, so maybe the target branch should be a new one called `6.x` or something like that, instead of `main`.

fritz-c:
I only spotted a couple other breaking changes that had been proposed. There weren't any in GitHub issues, but there were a couple `@deprecated`-marked exports in the code that we could get rid of for v6.

fritz-c (approved):

PR #984, opened 2022-11-12
fix: only add tabIndex when keyboardControls enabled

### LINKED ISSUE #982: Slider frame tabIndex accessibility issues
We've recently had an external accessibility audit on a site built with lots of nuka-carousel. 

An issue they're currently failing us on is that the slider frame container has `tabIndex=0` applied to it - which is considered a fail as it enables the element to receive focus without being interactive.

That sounds logical - but does anyone know why tabIndex was added to that element?

parkersweb:
### Description

Prevents `tabIndex=0` being added to slider-frame when keyboard controls are not enabled. This fixes an accessibility issue whereby the slider-frame was previously focussable even though it was non-interactive when `enableKeyboardControls=false`.

Not sure if this requires additional tests?

Fixes #982

#### Type of Change

### How Has This Been Tested?

Verified in nextJs sample app that the `tabindex=0` attribute is only present when `enableKeyboardControls=true`.

### DIFF
--- a/packages/nuka/src/carousel.tsx
+++ b/packages/nuka/src/carousel.tsx
@@ -646,7 +646,7 @@ export const Carousel = (rawProps: CarouselProps): React.ReactElement => {
         }}
         aria-label={frameAriaLabel}
         role="region"
-        tabIndex={enableKeyboardControls ? 0 : undefined}
+        tabIndex={enableKeyboardControls ? 0 : -1}
         onKeyDown={enableKeyboardControls ? onKeyDown : undefined}
         ref={carouselRef}
         onMouseUp={onMouseUp}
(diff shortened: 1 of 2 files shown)


packages/nuka/src/carousel.tsx
```diff
@@ -646,7 +646,7 @@ export const Carousel = (rawProps: CarouselProps): React.ReactElement => {
         }}
         aria-label={frameAriaLabel}
         role="region"
-        tabIndex={0}
+        tabIndex={enableKeyboardControls ? 0 : undefined}
```
fritz-c, line 649:
```suggestion
        tabIndex={enableKeyboardControls ? 0 : -1}
```

Let's fall back to -1, just so we don't break any existing logic that programmatically focuses on this component.

fritz-c:
Also, there's this item from the PR template checklist that you skipped:
> I have recorded any user-facing fixes or changes with `pnpm changeset`.

That will help us make the next release with your changes.

parkersweb:
thanks @fritz-c  - should be done now!

fritz-c (approved):

PR #1012, opened 2023-04-07
Add Cypress tests plus GitHub workflow back to v6

carloskelly13:
### Description

- Adds back Cypress tests that use the demos embedded in the new doc's website to test a standard and wrap-around carousel
- Adds Cypress tests for navigational touch gestures with different speeds and distance values 
- Fixes a bug that makes sure Nuka is built upon `pnpm install` of the doc's website to ensure dependencies are found

#### Type of Change

### How Has This Been Tested?

This has been tested locally by running the Cypress tests and in the CI workflow via the passing GitHub action.


website/package.json
```diff
@@ -4,6 +4,7 @@
   "private": true,
   "scripts": {
     "docusaurus": "docusaurus",
+    "postinstall": "pnpm run -C ../ build",
```
carloskelly13, line 7: `tsup` is pretty fast and I think running this on `postinstall` is pretty harmless since these docs need Nuka built to start up and run.

gksander: Seems reasonable to me.

gksander (approved):
lgtm

PR #1015, opened 2023-04-25
Setup package provenance

gksander:
Sets up NPM package provenance.


carloskelly13:
Let's add the copy README task and then chat about bumping our peer dependency on the library to React 18.

gksander:
@carloskelly13 the README is already being copied, you had already set that up once upon a time I believe! But yes, let's chat React 18 separately and figure out what the play is there.

carloskelly13 (approved):
Oh yeah! Sounds great.

Issue #1001, opened 2023-03-20
Attributes do not match their roles on carousel dots


LinnJS:
# Bugs and Questions

### Prerequisites

- [X] I've searched open [issues](https://github.com/FormidableLabs/nuka-carousel/issues) to make sure I'm not opening a duplicate issue
- [X] I have read through the docs before asking a question
- [X] I am using the latest version of nuka-carousel

### Describe Your Environment

- What version of nuka-carousel are you using? 
"nuka-carousel": "^5.5.1",
- What version of React are you using? 
"react": "18.2.0",
- What browser are you using? 
 Brave Browser v1.49.120

### Describe the Problem
We are using the default `nuka-carousel` carousel dots and styling them by overwriting styles in a css file. When running a lighthouse audit our accessibility score dropped from a `100` to a `90` when implementing a carousel with dots. On our `100` previous score we had a `nuka-carousel` but used custom prev next buttons with no issues. Adding another `nuka-carousel` with dots dropped the score and caused an accessibility error.

Error:
Each ARIA `role` supports a specific subset of `aria-*` attributes. Mismatching these invalidates the `aria-*` attributes. [Learn more](https://web.dev/aria-allowed-attr/?utm_source=lighthouse&utm_medium=devtools).

![image](https://user-images.githubusercontent.com/23741323/226388797-4c90facc-5e30-4194-932b-56aa751c16dd.png)

#### Expected behavior:
Aria tags should match roles of element

#### Actual behavior
Aria error of ARIA `role` supports a specific subset of `aria-*` attributes. Mismatching these invalidates the `aria-*` attributes. [Learn more](https://web.dev/aria-allowed-attr/?utm_source=lighthouse&utm_medium=devtools).

Feel free to go to website and run lighthouse with accessibility audit checked.

## Website link
https://bunkr-p9eavz7e8-linnjs.vercel.app/

LinnJS:
This looks to have been fixed thank you @clarkgunn

PR #1031, opened 2023-09-14
Update accessible roles for v5

clarkgunn:
### Description

Adjusts roles and attributes to better align with the WCAG Aria Authoring Practices Guide https://www.w3.org/WAI/ARIA/apg/patterns/carousel

* New `tabbed` prop determines whether roles should follow the Basic Carousel or Tabbed Carousel requirements.
* New `landmark` prop determines whether role should be a `region` landmark or have role `group`. Role and label have been moved from the slider frame to the parent carousel element so controls are contained within the label and role. Landmark regions should be intentional, so the default value is set to false.
* New `carouselId` prop gives the carousel frame an `id` attribute which can be uniquely targeted by controls where there might be multiple carousels on a page. This could be replaced internally with `useId` with React 18.
* Default `frameAriaLabel` has been changed to a slightly more useful 'Slider', the role description of 'carousel'  will already be announced 
* Default pagination dots have been given the role `tab`, and `aria-controls` targeting their slide.
* Pagination dots parent has been given the role `tablist`. As `tab`s need to be the immediate child of `tablist`, and since `tablist` already has list semantics, `ul` and `li` have been removed. This should resolve https://github.com/FormidableLabs/nuka-carousel/issues/1001 and the issue raised in https://github.com/FormidableLabs/nuka-carousel/pull/803 regarding `aria-selected`.
* When `tabbed` is true, slides will have the role `tabpanel` indicating they are the target of the `tab`, otherwise they will have the role `group` and a description of `slide`.
* When `tabbed` is false, default pagination dots will not be rendered.
* Default next and previous buttons now target the slider frame with `aria-controls`.

Adds dev dependencies: `jest-axe`, `@types/jest-axe`

Fixes https://github.com/FormidableLabs/nuka-carousel/issues/1001

#### Type of Change

These changes will impact the roles and attributes when default props are used. I believe this will improve accessibility but may be considered a breaking change.

### How Has This Been Tested?

New jest and cypress tests were added. Existing cypress tests needed wait logic to pass, akin to this change https://github.com/FormidableLabs/nuka-carousel/pull/887/commits/1f7f0200e5b5fe45f0f55379d4c909836fc2d5cb

Cypress and Jest tests and Lighthouse accessibility report should all pass. Attributes and roles can be verified in devtools.

A tabbed carousel with chrome devtools displaying accessible roles and attributes.

An untabbed region landmark carousel with chrome devtools displaying accessible roles and attributes.

### Checklist


packages/nuka/src/carousel.test.tsx
```diff
@@ -140,39 +158,45 @@ describe('Carousel', () => {
...
       keyCodeConfig,
       slideCount,
       beforeSlide,
+      frameAriaLabel: 'keyboard',
+      landmark: true,
+      carouselId,
     });
 
-    const carouselFrame = screen.getByRole('region');
+    const sliderFrame = screen.getByTestId(`${carouselId}-slider-frame`);
```
clarkgunn, line 171: `frameAriaLabel` now references the whole carousel, this ensures the same slider frame element is being selected.

packages/nuka/src/types.ts
```diff
@@ -222,6 +225,11 @@ export interface InternalCarouselProps {
    */
   beforeSlide: (currentSlideIndex: number, endSlideIndex: number) => void;
 
+  /**
+   * Unique id attribute for the carousel which may be referenced by aria attributes.
+   */
+  carouselId: string;
```
clarkgunn, line 231: I went with `carouseld` to disambiguate between `id` attributes and props on other elements.

packages/nuka/src/types.ts
```diff
@@ -428,6 +441,11 @@ export interface InternalCarouselProps {
    */
   swiping: boolean;
 
+  /**
+   * Whether tab pagination is used to set appropriate roles for slides.
+   */
+  tabbed: boolean;
```
clarkgunn, line 447: `tabbed` doesn't exactly align with the "pagination" and "dots" verbiage currently used, but it aligns with WCAG terminology.

packages/nuka/cypress/integration/carousel.spec.ts
```diff
@@ -1,4 +1,5 @@
 /* eslint-disable cypress/no-unnecessary-waiting */
+const waitThreshold = 300;
```
clarkgunn, line 2: http://localhost:3000/?slides=4&params=%7B%22landmark%22:true,%22tabbed%22:false%7D

packages/nuka/src/default-controls.tsx
```diff
@@ -268,8 +274,15 @@ export const PagingDots = ({
...
   const currentSlideBounded = getBoundedIndex(currentSlide, slideCount);
 
+  if (!tabbed) return null;
+
   return (
-    <ul className={pagingDotsContainerClassName} style={listStyles}>
+    <div
+      className={pagingDotsContainerClassName}
+      style={listStyles}
+      aria-label="Choose slide to display."
```
carloskelly13, line 283: Should we let this be customizable with this string as default? (maybe this is more for an eventual port of this for v6 and not this specific use case)

carloskelly13 (approved):
Thank you for these improvements and getting this more inline. This will be also great to forward-port to v6 as well.

PR #1043, opened 2024-02-01
v8 Rewrite

melvin-chen:
Nuka v8
---

Nuka v8 is an upgrade that aims to simplify the API and align the carousel with modern alternatives. 

### Highlights
- Size reduced by more than 85% of the original package size
- Much more responsive, don’t have to render two carousels and hide one with CSS for each breakpoint
- Animations aren’t manually done and rely and browser/OS animations that are much more accessible, touch friendly, and result in a smaller bundle size
- You can have different widths for each slide if needed
- Voiceover accessibility is more intuitive for keyboard navigation users.
- Documentation is improved with more in-depth examples and Typescript typing

### Breaking Changes

- Removed many props (with reason!). This is documented in the [v8 Upgrade Guide page](https://nuka-carousel-website-git-v8-rewrite-formidable-labs.vercel.app/v8-upgrade-guide)
- Wrap around is not the same as it was before where the carousel would seamlessly flow from the end to the start again. It currently just scrolls backward to the first slide again.


.github/workflows/cypress.yml
melvin-chen, line 1: Removing this file since:
1. It didn't run cypress tests to begin with, just the normal jest tests
2. I'm favoring the built in Storybook tests for tests that require a DOM. (More on this in the storybook test file)

packages/nuka/cypress/e2e/standard-carousel.cy.ts
melvin-chen, line 1: These are mostly covered in the storybook interaction tests

packages/nuka/cypress/e2e/swipe-gestures-carousel.cy.ts
melvin-chen, line 1: These should be native to the OS now. We shouldn't be manually calculating the swipe gestures to navigate around, and since these are native, we shouldn't have to test it.

packages/nuka/cypress/e2e/wrap-around-carousel.cy.ts
melvin-chen, line 1: The tests in this file are actually pretty useful, but since the rewrite changes how we handle wrapping, we should come back to this later.

packages/nuka/src/Carousel/CarouselTests.stories.tsx
melvin-chen, line 1: I want the tests to live in storybook through interaction and its built in Cypress testing if needed, but I feel like this file is really messy and it's hard to tell what's being tested without describe() and it(). Really could use some pointers here.

docs/api/options.mdx
```diff
@@ -0,0 +1,118 @@
...
+### Details
+
+| Prop Name          | Type    | Default Value |
+| :----------------- | :------ | :------------ |
+| `autoplay`         | boolean | `false`       |
+| `autoplayInterval` | number  | `3000`        |
+
+### Example
+
+<BasicDemo className="standard-demo" autoplay={true} />
```
carbonrobot, line 24: The basic demo appears to always have two prev/next buttons displayed. Does carousel always display these buttons? Or, if that is for the demo alone, perhaps we should omit them to avoid confusion.

docs/api/options.mdx
```diff
@@ -0,0 +1,118 @@
...
+
+### Details
+
+| Prop Name        | Type                          | Default Value |
+| :--------------- | :---------------------------- | :------------ |
+| `scrollDistance` | "slide" \| number \| "screen" | `slide`       |
+
+Using a number will cause the carousel to scroll that many pixels each time.
+Using "slide" will advance the carousel one slide each time.
+Using "screen" will advance the carousel by the width of the visible carousel.
```
carbonrobot, line 40: Screen appears to have a bug that can be reproduced by scrolling all the way to the end, then scrolling back one time.

docs/api/options.mdx
```diff
@@ -0,0 +1,118 @@
...
+  showPageIndicators={true}
+  pageIndicatorProps={{
+    containerClassName: 'flex items-center justify-center py-4 gap-2',
+    pageIndicatorClassName:
+      'w-3 h-3 p-0 rounded-full bg-gray-200 border-none cursor-pointer hover:bg-pink-200',
+    currentPageIndicatorClassName: 'bg-pink-500 hover:bg-pink-500',
+  }}
+  scrollDistance="screen"
+/>
+
```
carbonrobot, line 101: We should include code examples that are visible on the page as well

docs/api/options.mdx
```diff
@@ -0,0 +1,118 @@
...
+### Details
+
+| Prop Name          | Type   | Default Value |
+| :----------------- | :----- | :------------ |
+| `wrapperClassName` | string | `undefined`   |
+
+### Example
+
+<BasicDemo className="standard-demo" wrapperClassName="gap-4" />
```
carbonrobot, line 118: Without showing the actual code used in the demo, this example is confusing. The description talks about applying a style to the wrapper, and although thats what we did here, we used `gap-4` which alters the children of the wrapper which may confuse people into thinking we meant the wrapper of the slide and not of the carousel.

I recommend we consider an alternative name for this property that better describes what it will apply to. Consider something like `carouselClassName` or perhaps just the native `className`

docs/v8-upgrade-guide.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+`cellSpacing` - this should just be controlled through passed CSS.
+
+`defaultControlsConfig`
+
+`disableAnimation` - should just be controlled through passed CSS.
+
+`dragging` - this is always on, it defaults to the OS/browser settings.
+
+`dragThreshold` - this defaults to the OS/browser settings.
```
carbonrobot, line 45: We may want to keep the drag properties available to override regardless of OS settings

docs/api/options.mdx
carbonrobot: Touch scrolling on mobile devices appears to ignore this setting entirely

docs/api/options.mdx
```diff
@@ -0,0 +1,118 @@
...
+    <p>
+      Scroll by the width of the "screen" or the container the carousel is in.
+    </p>
+    <BasicDemo className="screen=distance-demo" scrollDistance="screen" />
+  </TabItem>
+</Tabs>
+
+---
+
+## Page Indicators
```
carbonrobot, line 69: On mobile devices, clicking the "middle" indicator on the example scrolls to the center of the slides incorrectly.

image

docs/v8-upgrade-guide.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+`easing` - should just be controlled through passed CSS.
+
+`edgeEasing` - should just be controlled through passed CSS.
+
+`enableKeyboardControls` - this should be native without interference. If the content is focusable through normal tabbing, it will focus.
```
carbonrobot, line 51: v8 appears to ignore the Scroll setting and the keyboard scrolls 1px for each keyboard button press.

docs/v8-upgrade-guide.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+`keyCodeConfig` - should be native.
+
+`landmark`
```
carbonrobot, line 55: We appear to have lost all accessibility indicators in v8

docs/api/options.mdx
carbonrobot: The slide option does not appear to show the final slide in its entirety before wrapping around
slides

packages/nuka/package.json
```diff
@@ -89,7 +93,7 @@
     "url": "https://github.com/FormidableLabs/nuka-carousel/issues"
   },
   "homepage": "https://github.com/FormidableLabs/nuka-carousel",
-  "sideEffects": false,
+  "sideEffects": true,
```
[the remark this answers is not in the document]
melvin-chen: I switched this to `true` to get the build to include the CSS file and the Carousel file to import the CSS file correctly. The build won't break with sideEffects: false, but it won't have any of the styles applied.

packages/nuka/src/Carousel/ExampleSlide.tsx
```diff
@@ -0,0 +1,40 @@
...
+    style={{
+      ...getExampleSlideStyles(index),
+      flexDirection: 'column',
+    }}
+    onFocus={(event) => event.target.scrollIntoView()}
+    className="focusable"
+  >
+    Card {index}
+  </a>
+);
```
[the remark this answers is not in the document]
melvin-chen: From what I understand from the tsup config, it should only package the exports from `src/Carousel/index.tsx`, which would be just the Carousel and SlideHandle. I tested this by comparing the bundle size after deleting the story/style/test files and deleting the content from Carousel and Slide Handle, and the bundle was only affected when deleting Carousel and Slide Handle.

docs/api/options.mdx
melvin-chen: I think this is from the width of the window changing, or that was the only way I could reproduce this. I've added a resize listener that will update these indicators and where the indicators go when clicked.

docs/api/options.mdx
melvin-chen: I think this might be the same issue as the indicator issue; I'm guessing you changed the window size to simulate mobile and the component didn't recalculate its number of slides relative to the screen size. This should be solved with the resize listener in 3f4df56

docs/v8-upgrade-guide.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+`tabbed` - should be native.
+
+`withoutControls` - controls are not rendered by default. Use methods to build your own next/prev triggers or style the page controls with `showPageIndicators` and `pageIndicatorProps`.
+
+`zoomScale` - should just be controlled through passed CSS.
+
+---
+
+## Props in consideration
```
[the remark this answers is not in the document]
melvin-chen: Just added some more context for these

docs/api/options.mdx
carbonrobot: Looks good now. Touch still has some weirdness where it ignores the setting

docs/api/options.mdx
carbonrobot: I tried a reload in a device simulator, here is what I'm seeing on Android Chrome.

slides_2

melvin-chen: oh interesting, yeah just got it to reproduce on a physical device. ill look into how the last page is being calculated

packages/nuka/src/Carousel/Carousel.test.tsx
```diff
@@ -0,0 +1,27 @@
...
+  // it.todo('passes the same className for each of the children');
+  it.todo('is draggable through touch controls');
+  // it.todo('has a next and previous function that is callable from its parent');
+  // it.todo('can slide');
+  it.todo('omits slides whose children are falsy');
+
+  // it.todo('autoplays at the right rate');
+
+  it.todo('returns the right number of page dots');
+});
```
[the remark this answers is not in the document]
melvin-chen: Mentioned briefly in this self-comment https://github.com/FormidableLabs/nuka-carousel/pull/1043#discussion_r1503457579 . Eventually, I'd want to get rid of this test file since it won't be run after removing the cypress.yml file.

docs/api/options.mdx
melvin-chen: Added code examples into both options and methods docs!

docs/api/options.mdx
melvin-chen: Removed the buttons from the carousel examples that don't really apply. I also wrote in the v8 Upgrade guide doc under Props Under Consideration to add default or built-in buttons.

docs/api/options.mdx
melvin-chen: This should be fixed in [c2e59d2](https://github.com/FormidableLabs/nuka-carousel/pull/1043/commits/c2e59d29d439468809c4b4b865543a7c3645ced9), the carousel seems to stop just short of the final indicator on some mobile devices. I added a Math.round() to the page indicator checker to just round to the closest indicator that the carousel has passed.

docs/api/options.mdx
melvin-chen: There was a different bug similar to this on mobile devices that should be fixed in c2e59d2. I rounded the manual scrolling check since something was pushing the scroll indices to be some decimal point off, which would cause some slides to skip when pressing next or not move when pressing back.

PR #1064, opened 2024-06-20
Fix for leftover width less than half of container/screen size

jduthon:
### Description

During implementation of the carousel, we noticed sometimes a weird behavior where the amount that can be scrolled left does not go all the way to the end of the content.

See example from this before recording

https://github.com/FormidableLabs/nuka-carousel/assets/4168055/b7fc3495-ec21-4c69-a9d4-f9c5213f176b

You can notice there the scroll stops at the Slide 8, but technically there are 10 slides (so the last slide is numbered 9)

This is due to the number of pages being calculated as a rounded value, hence if the leftover width is less than half of the container size, the number of pages will be rounded down to the lower value and it'll be impossible to scroll to the end with the buttons/scroll indicators.

I propose the following fix:
1. Use a ceiled value instead of the rounded one, so it'll always include the last missing leftover part
2. I calculated the last scroll value as a modulo of the width so that the scroll value is precise. It's not technically mandatory (at least for Chrome) as the browser only scrolls to the available content, but I thought if this value is ever used for some other computations it might make sense to have it be precise

Here is the after result from the same story at the same screen size:

https://github.com/FormidableLabs/nuka-carousel/assets/4168055/85933140-837b-44f3-9e25-d7424c318fd1

#### Type of Change

- N/A New feature (non-breaking change which adds functionality)
- N/A Breaking change (fix or feature that would cause existing functionality to not work as expected)
- N/A This change requires a documentation update

### How Has This Been Tested?

- Covered with unit tests
- Tested on the storybook examples

### Checklist

- [X] My code follows the style guidelines of this project (I have run `pnpm run lint`)
- [X] New and existing unit tests pass locally with my changes (I have run `pnpm run test:ci-with-server`/`pnpm run test`)
- N/A I have made corresponding changes to the documentation

### DIFF
--- a/packages/nuka/src/hooks/use-measurement.test.tsx
+++ b/packages/nuka/src/hooks/use-measurement.test.tsx
@@ -121,7 +121,7 @@ describe('useMeasurement', () => {
     const { totalPages, scrollOffset } = result.current;
 
     expect(totalPages).toBe(2);
-    expect(scrollOffset).toEqual([0, 500]);
+    expect(scrollOffset).toEqual([0, 400]);
   });
 
   it('should return measurements for screen with fractional pixels', () => {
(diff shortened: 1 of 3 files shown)


packages/nuka/src/hooks/use-measurement.test.tsx
```diff
@@ -150,8 +150,36 @@ describe('useMeasurement', () => {
 
     const { totalPages, scrollOffset } = result.current;
 
-    expect(totalPages).toBe(3);
-    expect(scrollOffset).toEqual([0, 573, 1146]);
+    expect(totalPages).toBe(4);
+    // 573 * 0, 573 * 1, 573 * 2, 573 * 3 + (1720 - 573 * 3)
```
jduthon, line 154: Not sure if this is useful, but I thought a code comment explaining where the values are coming from for the `toEqual` might be helpful?
Happy to edit or delete if it adds more confusion for now

packages/nuka/src/hooks/use-measurement.test.tsx
```diff
@@ -150,8 +150,36 @@ describe('useMeasurement', () => {
...
+    expect(scrollOffset).toEqual([0, 573, 1146, 1147]);
+  });
+
+  it('should return measurements for screen with less than half offset', () => {
+    const element = {
+      current: {
+        // this test covers that even when the leftover width is less than
+        // half of the screen width, it should still be scrollable so that user can see
+        // the small overflow
+        scrollWidth: 600,
```
jduthon, line 164: New test that verifies the fix works correctly

PR #1075, opened 2024-10-07
Improve the stability of swiping

carbonrobot:
Improve the stability of swiping by using a new `minSwipeDistance` property and calculating the swipe direction to determine the users intent.

### Current

nuka_main

### Fix

nuka_branch


mwritter (approved):
lgtm

PR #1077, opened 2024-10-08
Improve the stability of swiping

carbonrobot:
Improve the stability of swiping by using a new `minSwipeDistance` property and calculating the swipe direction to determine the users intent. 

Previously we used an input delay (debounce) on drag to determine when to fire the page action. This had the unintended side effect of causing the carousel to "bounce" as the drag velocity did not always match the debounce duration.

This changeset now uses a swipe distance on drag to determine when to move to the next slide/screen, which is more predictable. This also sets up the code to handle mouse events in the future and possibly a vertical scroll option.

### Current

nuka_main

### This PR

nuka_branch


colinrobertbrooks (approved):
👍🏻 

Tested on iPhone: https://nuka-carousel-website-git-feat-stable-swipe-formidable-labs.vercel.app/open-source/nuka-carousel/docs/api/swiping

